PR #1, opened 2022-03-12
chore: add different types of buttons

nesrinesghaier:


package.json
```diff
@@ -68,6 +68,7 @@
     "react": "^17.0.2",
     "react-app-rewired": "^2.2.1",
     "react-dom": "^17.0.2",
+    "react-icons": "^4.3.1",
```
bacali95, line 71: This should be removed since it was added as a direct dependency

src/components/Button.tsx
```diff
@@ -11,11 +12,19 @@ type GradientDuoTone =
   | 'pinkToOrange'
   | 'tealToLime'
   | 'redToYellow';
-
+type Size = 'extraSmall' | 'small' | 'medium' | 'large' | 'extraLarge';
```
bacali95, line 15: Please use short names like:
```ts
type Size = 'xs' | 'sm' | ...
```

src/components/Button.tsx
```diff
@@ -11,11 +12,19 @@ type GradientDuoTone =
...
+type IconPosition = 'left' | 'right';
 export type ButtonProps = ComponentProps<'button'> & {
   pill?: boolean;
   outline?: boolean;
+  disabled?: boolean;
```
bacali95, line 20: I don't see why `disabled` should be mentioned since it is already included in the type `ComponentProps<'button'>`!

src/components/Button.tsx
```diff
@@ -11,11 +12,19 @@ type GradientDuoTone =
   | 'pinkToOrange'
   | 'tealToLime'
   | 'redToYellow';
-
+type Size = 'extraSmall' | 'small' | 'medium' | 'large' | 'extraLarge';
+type IconPosition = 'left' | 'right';
```
bacali95, line 16: I think it is more logical to have the position type as `'start' | 'end'`

src/components/Button.tsx
```diff
@@ -65,35 +74,74 @@ const gradientDuoToneClasses: Record<GradientDuoTone, string> = {
...
+  icon: Icon,
+  iconPosition,
   color = 'blue',
   gradientMonochrome,
   gradientDuoTone,
   ...props
 }) => {
   return (
     <button
+      disabled={loader || disabled}
```
bacali95, line 103: We should not force the button to be disabled if it is loader, we should let the disabled decision to the developers.

src/components/Button.tsx
```diff
@@ -65,35 +74,74 @@ const gradientDuoToneClasses: Record<GradientDuoTone, string> = {
...
       className={classNames(
-        'flex items-center justify-center p-0.5 text-sm text-center font-medium group',
+        'flex h-min items-center justify-center p-0.5 text-center font-medium group',
         pill ? 'rounded-full' : 'rounded-lg',
         !gradientMonochrome && !gradientDuoTone && colorClasses[color],
         !gradientDuoTone && gradientMonochrome && gradientMonochromeClasses[gradientMonochrome],
         gradientDuoTone && gradientDuoToneClasses[gradientDuoTone],
+        {
+          'bg-blue-400 dark:bg-blue-500 hover:bg-blue-400 dark:hover:bg-blue-400 cursor-not-allowed': disabled,
```
bacali95, line 111: What if the color is red or yellow? I think we can just reduce the opacity

src/components/Button.tsx
```diff
@@ -65,35 +74,74 @@ const gradientDuoToneClasses: Record<GradientDuoTone, string> = {
...
+        },
       )}
       type="button"
       {...props}
     >
       <span
-        className={classNames('px-5 py-2.5 rounded-md', {
+        className={classNames('flex', size && sizeClasses[size], {
```
bacali95, line 118: Looking at the signature of the component the `size` property is always defined since it has the default value `medium`, so we can get rid of the `size &&` part

src/components/Button.tsx
```diff
@@ -65,35 +74,74 @@ const gradientDuoToneClasses: Record<GradientDuoTone, string> = {
...
+          'rounded-full': outline && pill,
+          'text-sm px-2.5': iconButton,
+          'gap-2': loader,
         })}
       >
+        {loader && <Spinner size="xs" />}
+
+        {iconPosition === 'left' && Icon && <Icon className="mr-1.5 w-5 h-5" />}
         {children}
+        {iconPosition === 'right' && Icon && <Icon className="ml-1.5 w-5 h-5" />}
```
bacali95, line 131: If the button is an `IconButton` this check and the above one should not happen

src/components/Button.tsx
```diff
@@ -65,35 +74,74 @@ const gradientDuoToneClasses: Record<GradientDuoTone, string> = {
...
+        className={classNames('flex', size && sizeClasses[size], {
           'text-gray-900 transition-all ease-in duration-75 bg-white dark:bg-gray-900 group-hover:bg-opacity-0 group-hover:text-inherit dark:text-white':
             outline,
+          'rounded-md': outline && !pill,
+          'rounded-full': outline && pill,
+          'text-sm px-2.5': iconButton,
+          'gap-2': loader,
         })}
       >
+        {loader && <Spinner size="xs" />}
```
bacali95, line 127: I think the size property can be used here to determine the size of the loader to be the same as the button

src/components/Button.tsx
```diff
@@ -65,35 +74,74 @@ const gradientDuoToneClasses: Record<GradientDuoTone, string> = {
...
         {children}
+        {iconPosition === 'right' && Icon && <Icon className="ml-1.5 w-5 h-5" />}
+        {iconButton && Icon && <Icon className="w-5 h-5" />}
       </span>
+
+      {label && (
+        <span
+          className={classNames({
+            'inline-flex justify-center items-center mr-4 -ml-2 w-4 h-4 text-xs font-semibold text-blue-800 bg-blue-200 rounded-full':
+              !!label,
```
bacali95, line 139: If this is rendered, that means the label is already defined so we can use the class names directly without this check.

bacali95 (changes requested):

bacali95 (approved):

PR #19, opened 2022-04-11
Adding Storybook

rluders:
This PR adds the Storybook to the project.

> Storybook is an open source tool for building UI components and pages in isolation. It streamlines UI development, testing, and documentation.

It would simplify the current documentation as well to introduce some interesting tooling to test the components.

We could also apply some [theme](https://storybook.js.org/docs/react/configure/theming) to customize the look and feel of the storybook page.


package.json
```diff
@@ -81,7 +95,8 @@
     "react-syntax-highlighter": "^15.5.0",
     "standard-version": "^9.3.2",
     "tailwindcss": "^3.0.23",
-    "typescript": "^4.6.3"
+    "typescript": "^4.6.3",
+    "webpack": "5"
```
bacali95, line 99: why do we need `webpack` as a dev dependency?

src/components/Button/index.tsx
```diff
@@ -134,7 +134,7 @@ export const Button: FC<ButtonProps> = ({
           [iconSizeClasses[size]]: !!Icon,
         })}
       >
-        {Icon ? <Icon className="h-5 w-5" /> : children}
+        {Icon ? <Icon className={`h-5 w-5`} /> : children}
```
bacali95, line 137: Why this change?

src/components/Button/Button.stories.tsx
```diff
@@ -0,0 +1,17 @@
+import { Meta, Story } from '@storybook/react/types-6-0';
```
bacali95, line 1: I think we should move the stories next to the stories folder so they don't get built and shipped with the library.

bacali95 (changes requested):

package.json
rluders: It is overwriting some indirect dependency that includes the webpack 4. Storyboard just advised me do do it in order to avoid some issues.

src/components/Button/index.tsx
rluders: Ops! Probably some autofix of my IDE, but it doesn't make any sense.

src/components/Button/Button.stories.tsx
bacali95: Actually, since we use `tsc` to build the whole `src/components` folder all stories will be included as well, I just pushed a commit to ignore the `**/**.stories.tsx` files and by that we can keep using the best practices you just mentioned :grin:

PR #22, opened 2022-04-13
Improve documentation structure

rluders:
This PR should reduce the number of places to change when adding a component to the website documentation, by centralizing it in one single place.

With this PR, you only will need to change the file `routes.tsx` in order to:

- Add its page to the router
- Add it to the sidebar
- Create the component card at the dashboard page

### DIFF
--- a/src/Root.tsx
+++ b/src/Root.tsx
@@ -1,162 +1,14 @@
 import { FC, Suspense, useState } from 'react';
-import {
-  HiAnnotation,
-  HiArrowCircleDown,
-  HiBadgeCheck,
-  HiBell,
-  HiChevronDoubleRight,
-  HiClipboardList,
-  HiCollection,
-  HiCreditCard,
-  HiDeviceTablet,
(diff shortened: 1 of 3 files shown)


src/Root.tsx
```diff
@@ -191,7 +43,7 @@ export const Root: FC = () => {
         </div>
       </Navbar>
       <div className="flex h-full overflow-hidden bg-gray-50 dark:bg-gray-900">
-        <Sidebar collapsed={collapsed} itemsGroups={itemsGroups} />
+        <Sidebar collapsed={collapsed} itemsGroups={routes} />
```
bacali95, line 46: I think instead of having a two dimensions array just to satisfy this prop, we can have only one dimension array and use it as `itemGroups={[routes]}` in here.

bacali95 (changes requested):
I really liked the idea, just one minor change and we can merge it.

bacali95 (approved):

PR #23, opened 2022-04-13
Testing and Husky

rluders:
This PR fixes some issues related to husky, updates some tests dependencies, and last but not least introduces one basic unit test to the Button Component.

It also adds the test to the github action.


src/components/Button/tests.spec.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
```
bacali95, line 1: FYI using `react-jsx` in `tsconfig.json` allow us to not have this import in every jsx/tsx file.

rluders: Interesting. I had no idea about it.

Issue #10, opened 2022-03-29
Build Pagination component


zoltanszogyenyi:
Reference: https://flowbite.com/docs/components/pagination/

k2glyph:
https://github.com/themesberg/flowbite-react/pull/27

rluders:
@k2glyph add the issue number to the title, please. :)

PR #27, opened 2022-04-14
Pagination Component #10

k2glyph:
Please do review my PR @rluders @bacali95 and @mouracamila


src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,226 @@
+import classNames from 'classnames';
+import { FC, PropsWithChildren } from 'react';
+export type PaginationProps = PropsWithChildren<{
+  className?: string;
+  currentPage: number;
+  totalPages: number;
+  onPageChange: (page: number) => void;
+  // Optional props
+  showIcon?: boolean;
+  showIconOnly?: boolean;
```
rluders, line 10: Here is just a suggestion:

Instead of using `showIconOnly` and `showOnlyPreviousAndNext`, that I think can cause some confusion. I would prefer to have something like `displayFormat`, where you could pick between `pagination | navigation | navigation-group`. And them use the `showIcon` as a modifier to add the icon or not.

It probably would cover all the styles documented on original flowbite docs.

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+}) => {
+  // Logic for displaying page numbers
+  const pageNumbers = [];
+  const startPage = Math.max(1, currentPage - 3);
+  const endPage = Math.min(currentPage + 3, totalPages);
+  for (let i = startPage; i <= endPage; i++) {
+    pageNumbers.push(i);
+  }
+  if (tableDataPagination) {
+    return (
```
rluders, line 32: I'm seeing too many returns in the middle of the large block of codes, it kind of makes me confuse. I think that this is something that can be improved. Maybe you could create she inner components to handle those alternative templates?

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+export type PaginationProps = PropsWithChildren<{
+  className?: string;
+  currentPage: number;
+  totalPages: number;
+  onPageChange: (page: number) => void;
+  // Optional props
+  showIcon?: boolean;
+  showIconOnly?: boolean;
+  showOnlyPreviousAndNext?: boolean;
+  tableDataPagination?: boolean;
```
rluders, line 12: If you adopt my suggestion above, this will not be required anymore.

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,226 @@
...
+  showIcon,
+  showIconOnly,
+  showOnlyPreviousAndNext,
+  tableDataPagination,
+}) => {
+  // Logic for displaying page numbers
+  const pageNumbers = [];
+  const startPage = Math.max(1, currentPage - 3);
+  const endPage = Math.min(currentPage + 3, totalPages);
+  for (let i = startPage; i <= endPage; i++) {
```
rluders, line 28: I don't think that we need to run this loop for all the pagination styles. Do we?

rluders (changes requested):
In general looks good. I just made few comments that I think could help to improve some stuff.

src/components/Pagination/index.tsx
k2glyph: okay thanks @rluders. I will make this changes.

src/components/Pagination/index.tsx
k2glyph: This is for generating only few page numbers array based on the current page @rluders. 
For example:
current page =5, it will only display 2 3 4 5 6 7 8 page numbers for selection.

rluders: Yeah, I got it.

But, it will only make sense if you want to show those numbers, right?
If you are just showing the `Next` and `Prev` buttons... This loop doesn't make any sense... or did I got it wrong?

k2glyph: ooh yeah. got it @rluders.  Thanks

src/components/Pagination/index.tsx
k2glyph: done

src/components/Pagination/index.tsx
k2glyph: done

src/components/Pagination/index.tsx
k2glyph: done

src/components/Pagination/index.tsx
k2glyph: done

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+  showIcon: boolean,
+  displayFormat: 'pagination' | 'pagination-icon-only',
+) => {
+  const pageNumbers = [];
+  const startPage = Math.max(1, currentPage - 3);
+  const endPage = Math.min(currentPage + 3, totalPages);
+  for (let i = startPage; i <= endPage; i++) {
+    pageNumbers.push(i);
+  }
+  let showIconOnly = false;
```
rluders, line 41: These 3 lines (41-44) can be replaces with:

```ts
let showIconsOnly = displayFormat === 'pagination-icon-only';
```

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+}>;
+
+export const Pagination: FC<PaginationProps> = ({
+  currentPage,
+  totalPages,
+  onPageChange,
+  showIcon = false,
+  displayFormat = 'pagination',
+}) => {
+  if (displayFormat === 'pagination' || displayFormat === 'pagination-icon-only') {
```
rluders, line 20: Here is just a suggestion. You don't need to agree or change it if you don't want... But, I'm not a big fan of `if-then-else` statement.

There is two possible way to improve this part of the code:

First option: Use switch-case
```ts
switch (displayFormat) {
    case 'pagination':
    case 'pagination-icon-only':
        return renderPagination(currentPage, totalPages, onPageChange, showIcon);
    case 'navigation':
        return renderNavigation(currentPage, totalPages, onPageChange, showIcon);
    default:
        return renderNavigationGroup(currentPage, totalPages, onPageChange, showIcon);
}
```

Second option: Just use if, 'cause you are returning, so you don't need else-if
```ts
if (displayFormat === 'pagination' || displayFormat === 'pagination-icon-only') {
    return renderPagination(currentPage, totalPages, onPageChange, showIcon, displayFormat);
} 
if (displayFormat === 'navigation') {
    return renderNavigation(currentPage, totalPages, onPageChange, showIcon);
}
return renderNavigationGroup(currentPage, totalPages, onPageChange, showIcon);
```

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+                    fillRule="evenodd"
+                    d="M12.707 5.293a1 1 0 010 1.414L9.414 10l3.293 3.293a1 1 0 01-1.414 1.414l-4-4a1 1 0 010-1.414l4-4a1 1 0 011.414 0z"
+                    clipRule="evenodd"
+                  ></path>
+                </svg>
+                Previous
+              </>
+            )}
+            {showIconOnly && (
+              <svg className="h-5 w-5" fill="currentColor" viewBox="0 0 20 20" xmlns="http://www.w3.org/2000/svg">
```
rluders, line 70: Those SVG are being repeted? Can't it become a inner component?

rluders (commented):
Looks good. I'm just adding a few comments, most of them are just ways to improve the code, but... but in general I don't mind to get it merged right now and them work in some improves later.

What do you think @bacali95 @mouracamila?

src/components/Pagination/index.tsx
bacali95: We can even use the icons from 'react-icons' library since the original design itself already used one of them

src/components/Pagination/index.tsx
bacali95: I prefer the switch-case suggestion

k2glyph: Thanks. I have refactor the code and used the switch case.

src/components/Pagination/index.tsx
k2glyph: I have replace the svg with react-icons.

bacali95 (approved):

src/components/Pagination/index.tsx
```diff
@@ -0,0 +1,204 @@
...
+                  currentPage === page &&
+                    'bg-blue-50 py-2 px-3 text-blue-600 hover:bg-blue-100 hover:text-blue-700 dark:border-gray-700 dark:bg-gray-700 dark:text-white',
+                )}
+              >
+                {page}
+              </a>
+            </li>
+          );
+        })}
+
```
rluders, line 84: This blank line can be removed.

src/pages/PaginationPage.tsx
```diff
@@ -0,0 +1,87 @@
+import { FC, useState } from 'react';
+
+import { Pagination } from '../components';
+import { CodeExample, DemoPage } from './DemoPage';
+
+const PaginationPage: FC = () => {
+  const [currentPage, setCurrentPage] = useState(1);
+  const onPageChange = (page: number) => {
+    setCurrentPage(page);
+    console.log('Page changed to:', page);
```
rluders, line 10: No `console.log` please.

rluders (changes requested):
Still few things to change.

src/components/Pagination/index.tsx
k2glyph: okay @rluders

rluders (approved):
Thank you. 🚀

PR #32, opened 2022-04-16
feat: add tab component

nesrinesghaier:


src/components/tab/Tabs.tsx
```diff
@@ -0,0 +1,96 @@
...
+    notActive:
+      'bg-white hover:text-gray-700 hover:bg-gray-50 focus:ring-4 focus:ring-blue-300 focus:outline-none dark:hover:text-white dark:bg-gray-800 dark:hover:bg-gray-700',
+  },
+};
+
+export type TabsProps = PropsWithChildren<{
+  style?: TabTypes;
+}>;
+
+export const Tabs: FC<TabsProps> = ({ children, style = 'default' }) => {
```
bacali95, line 43: It would be better to have the same approach as the other components, like `Tab.Group` and `Tab.Item`

src/components/tab/Tabs.tsx
```diff
@@ -0,0 +1,96 @@
...
+            )}
+            onClick={() => setActiveTabIndex(index)}
+            disabled={tab.disabled}
+          >
+            {tab.icon && <tab.icon className="mr-2 h-5 w-5" />}
+            {tab.title}
+          </button>
+        ))}
+      </div>
+      <div className={tabs[activeTabIndex].className ? tabs[activeTabIndex].className : 'm-4'}>
```
bacali95, line 91: - can we have padding instead of margins.
- let's have like this `className={classNames('m-4', tabs[activeTabIndex].className)}` , aka not conditional

src/routes.tsx
```diff
@@ -256,6 +258,17 @@ export const routes: RouteProps[] = [
...
     },
   },
+  {
+    title: 'Tabs',
+    icon: MdTab,
+    href: '/tabs',
+    component: <TabsPage />,
+    group: false,
+    card: {
+      className: 'w-36',
```
bacali95, line 268: This one should be `w-64`

bacali95 (changes requested):
In general looks good, I just left some comments to address. It would be nice if you can add some stories and unit tests. Thanks :tulip:

rluders:
@nesrinesghaier any plans to continue with this one?

nesrinesghaier:
Will open a new one since force pushing is disabled

PR #61, opened 2022-04-21
feat: flowbite theme context provider

rluders:
This PR adds some theme context provider to handle splitting the part of the code that handles the dark mode and to be able to support some other theme features in the future.

**DISCLAIMER: THIS IS AN EXPERIMENT, WE ARE NOT SURE THAT IT BECOME PART OF FLOWBITE REACT YET.**

> This was heavily inspired by [windmill-react](https://github.com/estevanmaito/windmill-react-ui).


@types/index.d.ts
```diff
@@ -0,0 +1 @@
+type Mode = string | null;
```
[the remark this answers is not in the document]
bacali95: and also it is better to have a stronger type like `"light" | "dark" | "system"`

src/components/Flowbite/index.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC, PropsWithChildren, useLayoutEffect, useMemo } from 'react';
+import useDarkMode from '../../hooks/useDarkMode';
+import { ThemeContext } from '../../contexts/ThemeContext';
+
+export type FlowbiteProps = PropsWithChildren<{
+  children: any;
+  dark?: boolean;
+  usePreferences?: boolean;
```
bacali95, line 8: I think we should group the above two props because they are both related to the theme context.

src/components/Flowbite/index.tsx
```diff
@@ -0,0 +1,32 @@
+import { FC, PropsWithChildren, useLayoutEffect, useMemo } from 'react';
+import useDarkMode from '../../hooks/useDarkMode';
+import { ThemeContext } from '../../contexts/ThemeContext';
+
+export type FlowbiteProps = PropsWithChildren<{
+  children: any;
```
bacali95, line 6: we can remove this since it come automatically from `PropsWithChildren` type

src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+interface ThemeContextInterface {
+  mode?: Mode;
+  toggleMode?: any;
```
bacali95, line 5: Let's not use the type `any` anywhere and have always the exact type we need, for example here we need `() => void`

src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+interface ThemeContextInterface {
```
bacali95, line 3: Let's use `type` instead of `interface`

src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
bacali95, line 1: To keep the code consistent let's avoid this type of import and only get what we need.

src/hooks/useDarkMode.ts
```diff
@@ -0,0 +1,34 @@
+/* eslint-disable react-hooks/rules-of-hooks */
```
bacali95, line 1: If we are ignoring a rule that means there is something wrong with this file :)

src/components/Flowbite/index.tsx
```diff
@@ -0,0 +1,32 @@
...
+  useLayoutEffect(() => {
+    if (dark) {
+      if (setMode != null) {
+        setMode('dark');
+      }
+      document.documentElement.classList.add('dark');
+    }
+  }, [dark, setMode]);
+
+  const value = useMemo(
```
bacali95, line 23: We can rename this to `themeContextValue`

bacali95 (changes requested):

src/contexts/ThemeContext.tsx
rluders: `ThemeContextProps` sinde, I'm already exporting a `ThemeContext` const that is created by `createContext<>()`.

bacali95: The nice thing about typescript it allows you to export a type of a variable with the same name and it is smart enough to figure out wither the imported name (`ThemeContext`) is used as a value or a type.

src/contexts/ThemeContext.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+interface ThemeContextInterface {
+  mode?: Mode;
+  toggleMode?: any;
+}
+
+export const ThemeContext = React.createContext<ThemeContextInterface>({});
+
+interface ThemeProviderProps {
```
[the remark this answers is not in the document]
rluders: Actually, there is a even better option:

```ts
type ThemeProviderProps = ProviderProps<{}>;
```

src/hooks/useDarkMode.ts
rluders: Well, this is not always true. The rules sometimes are just recommendations or ways to check that you didn't forget something. :)

If you remove the comment and check it is just telling me that I'm early returning before some hooks... If I didn't make some mistake... Wich, I didn't. It is propositional.

src/hooks/useDarkMode.ts
```diff
@@ -0,0 +1,34 @@
...
+): [Mode, React.Dispatch<React.SetStateAction<Mode>> | null, (() => void) | null] => {
+  if (!usePreferences) return [null, null, null];
+
+  const [mode, setMode] = useState<Mode>(null);
+
+  const toggleMode = () => {
+    if (!mode) return;
+    const newMode = mode == 'light' ? 'dark' : 'light';
+    document.documentElement.className = '';
+    document.documentElement.classList.add(newMode);
```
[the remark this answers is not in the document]
rluders: hoho, didn't know about this one.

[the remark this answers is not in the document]
rluders: But it will change a lot of the behaviour. :thinking:

src/components/Flowbite/index.tsx
rluders: I'll change it, but I don't think that it will make any big difference to code readability. The variable scope context is very short. In these cases, short names are better.

Quick example, you don't need to say to a `createUser` function that an internal variable called `lastname` is `useLastName`, or `age` is `userAge`. :)

src/hooks/useDarkMode.ts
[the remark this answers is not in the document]
rluders: Yeah, I fact I could change it, but then it will not work as I'm expecting. I'll remove 3 lines here, and add a few more in other places. Don't see it as a :trophy:.

src/components/Flowbite/index.tsx
rluders: What would the benefits be?

src/contexts/ThemeContext.tsx
rluders: TS could be smart, but we are dumb sometimes. I will keep it explicit.

@types/index.d.ts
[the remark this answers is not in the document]
bacali95: That was just an example but for sure we can only have those two options.

src/components/Flowbite/index.tsx
bacali95: Later on, if we add more props to this component it might become unclear what every prop is doing.

src/hooks/useDarkMode.ts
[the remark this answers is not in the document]
bacali95: I don't see changing these lines as an optimization, I see a bug here that needs to be fixed, because if you do this affectation  `document.documentElement.className = '';` you might be breaking some behavior in the product.

src/contexts/ThemeContext.tsx
```diff
@@ -1,17 +1,16 @@
...
+};
 
-export const ThemeContext = React.createContext<ThemeContextInterface>({});
+export const ThemeContext = createContext<ThemeContextProps>({});
 
-interface ThemeProviderProps {
-  children: React.ReactNode;
-  value?: any;
-}
+type ThemeProviderProps = ProviderProps<{}>;
```
bacali95, line 12: Still we should replace `{}` with `ThemeContextProps`

bacali95 (changes requested):

rluders:
NOTE: I'm a little bit busy this week, but once I have some free time I'll continue work on this.

tulup-conner:
@rluders I am noticing a pattern with `flowbite-react` now that I'm using it. I create my own version of the `flowbite-react` component with `className` so I don't need to repeat it.

I have no idea if this is the best/ideal way to do this:

```js
const flowbiteTheme = {
  components: {
    Sidebar: "fixed overflow-auto top-16 h-screen lg:sticky lg:block"
  }
}
...
<Flowbite theme={flowbiteTheme}>
  ...
</Flowbite>
```

rluders:
> @rluders I am noticing a pattern with `flowbite-react` now that I'm using it. I create my own version of the `flowbite-react` component with `className` so I don't need to repeat it.
> 
> I have no idea if this is the best/ideal way to do this:
> 
> ```js
> const flowbiteTheme = {
>   components: {
>     Sidebar: "fixed overflow-auto top-16 h-screen lg:sticky lg:block"
>   }
> }
> ...
> <Flowbite theme={flowbiteTheme}>
>   ...
>   <Sidebar> // className inherits flowbiteTHeme.components.Sidebar
>     ..
>   </Sidebar>
> </Flowbite>
> ```

So, this is basically what I'm doing.

rluders:
There are still pieces to validate on this experiment, some results from code review, others from improvements and features that I want to include into it.

So it is probably far to be completed. However, feel free to leave any early feedback.

rluders:
@tulup-conner please, give it a try

src/lib/components/Flowbite/ThemeContext.tsx
```diff
@@ -0,0 +1,70 @@
...
+  return useContext(ThemeContext);
+}
+
+export const useThemeMode = (
+  usePreferences: boolean,
+): [Mode, React.Dispatch<React.SetStateAction<Mode>> | undefined, (() => void) | undefined] => {
+  if (!usePreferences) return [undefined, undefined, undefined];
+  const [mode, setMode] = useState<Mode>(undefined);
+
+  const savePreference = (m: string) => window.localStorage.setItem('theme', m);
```
tulup-conner, line 36: In theory you should need to check for `window` for all of these in SSR apps but I haven't noticed any issues in Remix

tulup-conner (commented):
This is so exciting 💯 I actually haven't added Accordions to my site yet LOL but I will now. In the meantime I am excited to test the rest of the components.

tsconfig.json
```diff
@@ -16,6 +16,6 @@
     "noEmit": true,
     "jsx": "react-jsx"
   },
-  "include": ["src"],
-  "exclude": ["node_modules"]
+  "include": ["src", "@types"],
+  "exclude": ["node_modules", "dist", "src/*.stories.ts", "src/*.test.ts"]
```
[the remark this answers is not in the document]
rluders: Ops! Nice catch. I just want to make sure that `tsc` isn't considering any unecessary stuff.

src/docs/pages/ThemePage.tsx
```diff
@@ -0,0 +1,67 @@
+import { FC } from 'react';
+import reactElementToJSXString from 'react-element-to-jsx-string';
+import { HiInformationCircle } from 'react-icons/hi';
+import { PrismLight as SyntaxHighlighter } from 'react-syntax-highlighter';
+import { dracula } from 'react-syntax-highlighter/dist/esm/styles/prism';
+import { Alert, Card, DarkThemeToggle } from '../../lib';
+import { Flowbite } from '../../lib/components/Flowbite';
+
+const ThemePage: FC = () => {
+  const SyntaxHighlighterFix = SyntaxHighlighter as any;
```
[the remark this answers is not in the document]
rluders: Just copy and pasted from DemoPage. LOL

src/lib/components/Flowbite/ThemeContext.tsx
rluders: well, I can access the localStorage directly, without use window...

PR #64, opened 2022-04-25
feat(component): Add `alt` attribute to `Avatar`s

tulup-conner:
Allows `Avatar`s to pass `alt` text to the underlying ``.

Now, by default, `Avatar`s will have `alt=""`, which browsers interpret appropriately as
[decorative images](https://www.w3.org/WAI/tutorials/images/decorative/).

To add `alt` text to your `Avatar`:

```js
<AvatarComponent
  alt="Default avatar with alt text"
  img="https://flowbite.com/docs/images/people/profile-picture-5.jpg"
/>
```

### DIFF
--- a/src/components/Avatar/index.tsx
+++ b/src/components/Avatar/index.tsx
@@ -4,6 +4,7 @@ import { AvatarGroup } from './AvatarGroup';
 import { AvatarGroupCounter } from './AvatarGroupCounter';
 
 export type AvatarProps = PropsWithChildren<{
+  alt?: string;
   size?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
   rounded?: boolean;
   bordered?: boolean;
@@ -36,6 +37,7 @@ const statusPositionClasses: Record<AvatarProps['statusPosition'] & string, stri
 };
 
 const AvatarComponent: FC<AvatarProps> = ({
+  alt = '',
   img,
   status,
   children,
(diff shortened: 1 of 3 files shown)


mouracamila:
Looks great to me :)

src/components/Avatar/index.tsx
```diff
@@ -57,7 +59,7 @@ const AvatarComponent: FC<AvatarProps> = ({
               'p-1': bordered,
             })}
             src={img}
-            alt="Rounded avatar"
+            alt={alt}
```
bacali95, line 62: Should we have a default alt value if none is specified?

tulup-conner: My logic is to default to "" which browsers interpret as a [decorative image](https://www.w3.org/WAI/tutorials/images/decorative/).

The difference is that screen readers will skip over it rather than reading aloud, "Rounded avatar".

IMO, it would be a good idea to not provide an alt text unless the developer has decided that image should have one, and what it should be.

But that is my opinion.

For example, if I use the Avatar in context -- like, next to the user's name whose Avatar it is -- I would probably see no need to have an alt text. It would just be redundant.

On the other hand, if I use the Avatar as a trigger for a Dropdown, it would be nice to provide alt text, which would provide the screen reader context for the Dropdown.

bacali95: Fair enough, I didn't see the `''` default :sweat_smile:

bacali95 (approved):

PR #65, opened 2022-04-25
feat(component): Add `imgAlt` attribute to `Card`s 

tulup-conner:
Allows `Card`s to pass `imgAlt` to provide `alt` text to the underlying ``, when an `imgSrc` has also been provided.

Now, by default, `Card`s with an image will have `alt=""`, which browsers interpret appropriately as [decorative images](https://www.w3.org/WAI/tutorials/images/decorative/).

To add `alt` text to your `Card` with an image:

```js
<Card
  imgAlt="Meaningful alt text for an image that is not purely decorative"
  imgSrc="https://flowbite.com/docs/images/people/profile-picture-5.jpg"
>
  ..
</Card>
```

Also renames the existing `Card with image` example to `Card with decorative image` to more clearly indicate `alt` text *should* be provided if the image is *not* purely decorative.

### DIFF
--- a/src/components/Card/index.tsx
+++ b/src/components/Card/index.tsx
@@ -2,13 +2,13 @@ import { FC, PropsWithChildren } from 'react';
 import classNames from 'classnames';
 
 export type CardProps = PropsWithChildren<{
-  alt?: string;
   className?: string;
   horizontal?: boolean;
+  imgAlt?: string;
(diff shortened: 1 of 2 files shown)


mouracamila:
Looks good to me

bacali95 (approved):
LGTM :+1:

PR #67, opened 2022-04-27
feat(build): Upgrade `@popperjs/*` to `@floating-ui/react-*`

### LINKED ISSUE #57: Tooltip Component
**Describe the bug**
The tooltip element is misaligned from the parent element 

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to ['storybook'](https://flowbite-react.com/storybook/?path=/story/components-tooltip--default-tooltip)
2. Click on Tooltip
3. See the error

**Expected behaviour**
The tooltip is aligned to parent element

**Screenshots**
image

I believe this is impact other components like the dropdown. 
image

**Desktop (please complete the following information):**
 - Browser: Chrome and Brave (Firefox is working)

**Smartphone (please complete the following information):**
n/a

**Additional context**
n/a

tulup-conner:
[Floating UI]() is the next evolution of [Popper](). 

We currently have a strange bug on Chromium causing `Dropdown`s/`Tooltip`s to be  misaligned with their target. ~~This upgrade happens to resolve #57, as a bonus!~~

Because `Tooltip` is used by `Dropdown`, both components have been upgraded.

There are no breaking changes to the API.

### DIFF
--- a/src/components/Tooltip/index.tsx
+++ b/src/components/Tooltip/index.tsx
@@ -1,86 +1,166 @@
-import { createPopper, Instance, Placement } from '@popperjs/core';
-import { FC, PropsWithChildren, ReactNode, useEffect, useRef, useState } from 'react';
+import { FC, PropsWithChildren, ReactNode, RefObject, useEffect, useRef, useState } from 'react';
 import classNames from 'classnames';
+import { arrow, autoPlacement, Placement, shift } from '@floating-ui/core';
+import {
+  autoUpdate,
+  flip,
+  Middleware,
+  offset,
+  useClick,
+  useFloating,
+  useFocus,
+  useHover,
+  useInteractions,
+  useRole,
+} from '@floating-ui/react-dom-interactions';
 
 export type TooltipProps = PropsWithChildren<{
   className?: string;
   content: ReactNode;
-  placement?: Placement;
+  placement?: 'auto' | Placement;
   trigger?: 'hover' | 'click';
   style?: 'dark' | 'light' | 'auto';
   animation?: false | `duration-${number}`;
   arrow?: boolean;
 }>;
 
+/**
+ * @see https://floating-ui.com/docs/react-dom-interactions
+ */
 export const Tooltip: FC<TooltipProps> = ({
+  animation = 'duration-300',
+  arrow = true,
   children,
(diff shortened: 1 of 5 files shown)


tulup-conner:
I still need to fix an issue where the tooltip arrow can obstruct content inside the tooltip itself.

tulup-conner:
Ugh, it looks like the bug is still present, but *only* in Storybook, and *only* in Chromium browsers. Because the issue is isolated to Storybook, I'm more inclined it's a bug on their end. Will look into that more.

src/components/Tooltip/index.tsx
```diff
@@ -1,86 +1,166 @@
...
+  placement,
+}: {
+  arrowRef: RefObject<HTMLDivElement>;
+  placement: 'auto' | Placement;
+}): Middleware[] {
+  const middleware = [];
+
+  middleware.push(offset(4));
+
+  if (placement === 'auto') {
```
rluders, line 140:
```ts
middleware.push(placement === 'auto' ? autoPlacement() : flip());
```

rluders (changes requested):

src/components/Tooltip/index.tsx
tulup-conner: Derp, change on the way

rluders:
@tulup-conner may I ask one 'related change'? Could you please change the decorator in this file:
https://github.com/themesberg/flowbite-react/blob/0b6bb5fbbaa160d02204d59d0c936e5572eba21f/.storybook/preview.js#L3

It needs to be something like this:
```ts
export const decorators = [
  (Story) => (
    <>
      <Style />
      <div className="flex h-screen justify-center items-center">
        <div>
          <Story />
        </div>
      </div>
    </>
  ),
];
```
The ideia is to centralize all components to the screen, so it would looks nice to preview it, and also 'cause I was having some "issues" (I know that it isn't an issue) to preview the placement to the tooptip when it when out of frame.

Otherwise it looks nice.

tulup-conner:
@rluders Done - I also added `<MemoryRouter>` as a wrapper because it is required for components that use `react-router-dom` `Link`. (Will be necessary to merge `Sidebar` stories in #69)

PR #68, opened 2022-04-27
chore: alert stories

mouracamila:
adding more story types to the alert component

### DIFF
--- a/src/components/Alert/Alert.stories.tsx
+++ b/src/components/Alert/Alert.stories.tsx
@@ -1,6 +1,7 @@
 import { Meta, Story } from '@storybook/react/types-6-0';
 
 import { Alert, AlertProps } from '.';
+import { HiInformationCircle, HiEye } from 'react-icons/hi';
 
 export default {
   title: 'Components/Alert',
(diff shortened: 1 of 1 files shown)


src/components/Alert/Alert.stories.tsx
```diff
@@ -1,6 +1,9 @@
 import { Meta, Story } from '@storybook/react/types-6-0';
 
+import React from 'react';
```
bacali95, line 3: Since react 17, this import is not required to be present anymore.

src/components/Alert/Alert.stories.tsx
```diff
@@ -1,6 +1,9 @@
...
+
 import { Alert, AlertProps } from '.';
+import { HiInformationCircle, HiEye } from 'react-icons/hi';
```
bacali95, line 6: Can we move this import under the storybook one?

bacali95 (changes requested):

src/components/Alert/Alert.stories.tsx
mouracamila: ok :)

src/components/Alert/Alert.stories.tsx
mouracamila: Sorry, I don't get it. What do you mean by that?

bacali95: I meant if we can move this line to line `2`

bacali95 (approved):

PR #74, opened 2022-05-02
chore(docs): Add `README`

tulup-conner:
This is a first draft for a README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,405 @@
-# :construction: flowbite-react :construction:
+<div align="center">
+  <h1>:construction: `flowbite-react` :construction:</h1>
+  <p>
+    <a href="https://flowbite.com">
+      <img alt="Flowbite - Tailwind CSS components" width="350" src="https://flowbite.s3.amazonaws.com/github/logo-github.png">
+    </a>
+    </p>
+    <p>
+    Build websites even faster with components on top of React and Tailwind CSS</p>
+</div>
+<p>
+    <a href="https://discord.com/invite/4eeurUVvTy"><img src="https://img.shields.io/discord/902911619032576090?color=%237289da&label=Discord" alt="Discord"></a>
+    <a href="https://www.npmjs.com/package/flowbite-react"><img src="https://img.shields.io/npm/dt/flowbite-react.svg" alt="Total Downloads"></a>
+    <a href="https://badge.fury.io/js/flowbite-react"><img alt="Latest release" src="https://badge.fury.io/js/flowbite-react.svg"></a>
+    <a href="https://flowbite.com/docs/getting-started/license/"><img src="https://img.shields.io/badge/license-MIT-blue" alt="Licenese"></a>
+</p>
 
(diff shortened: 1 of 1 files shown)


tulup-conner:
We need to add anchors on the sections of the Forms page that are linked in the new README Components section. That includes:

- [ ] File input
- [ ] Search input
- [ ] Select
- [ ] Textarea
- [ ] Checkbox
- [ ] Radio
- [ ] Toggle
- [ ] Range
- [ ] Floating label

tulup-conner:
@zoltanszogyenyi @rluders @bacali95 Let me know what you think!

tulup-conner:
Surprise! We also need to convert `HashRouter` to `BrowserRouter`, which I believe it already should've been, to allow anchors like `#range` on a page. This converts every `https://flowbite-react.com/#/...` to `https://flowbite-react.com/...`, which has the bonus of being more predictable.

src/index.tsx
```diff
@@ -1,14 +1,14 @@
...
+import { BrowserRouter } from 'react-router-dom';
 
 import './index.css';
 import 'flowbite';
 
 const container = document.getElementById('root')!;
 const root = createRoot(container);
 root.render(
-  <HashRouter>
+  <BrowserRouter>
```
bacali95, line 11: I used `HashRouter` because it is the only type of router that works with GitHub pages, but using the new domain I don't know if the `BrouwserRouter` will work or not :)

tulup-conner: Google tells me you just need to add a `basename` to the `Router`; I will deploy this to Pages on my own and get back.

bacali95: Let me give a shortcut: after deploying to Github actions try to consult one of the pages directly by typing `your-url/alerts` to the browser.

bacali95: First add `PUBLIC_URL` env to the `Build code 🛠` step in the `Publish Docs` workflow as `/flowbite-react` then pages will work for you

bacali95: Okay I might not seen this before but this can be a solution for the github 404 issue: https://create-react-app.dev/docs/deployment/#notes-on-client-side-routing

rluders (approved):
lgtm

src/index.tsx
tulup-conner: I think this is a dead end for now... I've remitted the router change from this PR so it can be merged.

README.md
```diff
@@ -1,5 +1,388 @@
...
+  <tr>
+    <td width="33.3333%">
+        <a href="https://flowbite.com/docs/forms/floating-label/">
+            <img alt="Tailwind CSS Floating Label" src="https://flowbite.s3.amazonaws.com/github/floating-label.jpg">
+        </a>
+    </td>
+  </tr>
+</table>
+
+## Getting started
```
bacali95, line 334: Can we move this section above the components section

bacali95 (changes requested):

README.md
tulup-conner: Done. I also moved the little intro from the Components section to the very top - thoughts?

bacali95: Nice, good enough for me

bacali95 (approved):

PR #76, opened 2022-05-02
Refactor code and move files to make library publishable again

bacali95:


rluders (approved):
Well... In general it lgtm... But I guess that I would prefer something a little bit different:
```
./docs/Root.tsx
./docs/pages/*
./src/components/*
```
But, up to you, since it is just preference... I think that what you did here is already an improvement... so, if you want to get it to merge it, go ahead.

bacali95:
> Well... In general it lgtm... But I guess that I would prefer something a little bit different:
> 
> ```
> ./docs/Root.tsx
> ./docs/pages/*
> ./src/components/*
> ```
> 
> But, up to you, since it is just preference... I think that what you did here is already an improvement... so, if you want to get it to merge it, go ahead.

This change is for including the helpers' folder to the build artifacts, I will move the docs as well

bacali95:
I cannot force push to this branch to apply the rebase will close and open a new one

PR #78, opened 2022-05-02
Refactor code and move files to make library publishable again

bacali95:


bacali95:
@tulup-conner I fixed some issues I find with the tooltip component (nice job btw) and added a feature to the sidebar item, it would be nice if you can take a look as well

src/lib/components/Sidebar/SidebarItem.tsx
```diff
@@ -23,8 +24,20 @@ const SidebarItem: FC<SidebarItemProps> = ({
 }) => {
   const { pathname } = useLocation();
 
-  return (
+  const Wrapper = ({ children: wrapperChildren }: PropsWithChildren<any>) => (
```
tulup-conner, line 27: `PropsWithChildren<Record<string, unknown>>` here over `any`

tulup-conner (approved):
looks good!

PR #85, opened 2022-05-05
feat(build): Add `cypress` for end-to-end (integration) testing

### LINKED ISSUE #35: Cypress + Axe testing
I believe that is important for us to have some automated tests. Since we do have the docs pages it should not be complicated to test the behaviors of the components. The test must cover usability and accessibility.

- Install Cypress
- Install Axe
- Create at least ONE example test

I don't know any alternative.

**Additional context**
The [Sumé LMS frontend](https://github.com/sumelms/frontend) already has those tests on it, maybe it would be easy to just bring them to this repository. Also, there is [this guide](https://www.deque.com/blog/how-to-test-for-accessibility-with-cypress/), that can be useful.

tulup-conner:
## Breaking changes

None.

## Features

- [x] `cypress` can now be used to create end-to-end (integration) tests in `cypress/integration` for pages in `src/docs/pages` (Resolves #35)
  - [x] `cypress-axe` can also be used to check for easy `a11y` fixes

## Bug fixes

None.

### Integration

- [x] Add integration tests for `Carousel` docs page


cypress/integration/docs/pages/CarouselPage.spec.ts
```diff
@@ -0,0 +1,86 @@
+describe('Docs / Carousel', function () {
+  beforeEach(function () {
```
rluders, line 2: I guess that we can also use arrow functions here.

cypress/support/commands.ts
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
rluders, line 1: Do we really need this file?

rluders (changes requested):

cypress/support/commands.ts
tulup-conner: Nope! In fact we dont need most of the files in cypress folder now. I was looking at setting up code coverage but it was more trouble than it's worth.

tulup-conner:
> @tulup-conner can we make it part of the CI?

I think we should but this won't currently work since there are several a11y issues on every Docs page. My thought is create another PR to resolve the outstanding site-wide a11y issues so the tests will pass.

rluders:
> > @tulup-conner can we make it part of the CI?
> 
> I think we should but this won't currently work since there are several a11y issues on every Docs page. My thought is create another PR to resolve the outstanding site-wide a11y issues so the tests will pass. But, I can roll that into this if you like.

Yep, it makes sense. We can enable it once we have the issues solved.

rluders (approved):

Issue #12, opened 2022-03-29
Build Timeline component


zoltanszogyenyi:
Reference: https://flowbite.com/docs/components/timeline/

joeychrys:
I can start this.
I'll structure it like:
`<Timeline>
	<Timeline.item>
		{content}
	<Timeline.Item>
</Timeline>`

rluders:
@joeychrys do you still interested in working on this one?

joeychrys:
> @joeychrys do you still interested in working on this one?

Hey, sorry. I probably can't get to it until Thursday next week. If this is holding you guys back, then by all means go ahead without me. I will have all the time in the world in like a week because I'll be done with school.

tulup-conner:
This can be closed now

PR #89, opened 2022-05-07
Timeline component

joeychrys:
Ref: This PR is related to issue #12


src/docs/pages/TimelinePage.tsx
```diff
@@ -0,0 +1,163 @@
+import { FC } from 'react';
+import { CodeExample, DemoPage } from './DemoPage';
+import { Timeline } from '../../lib';
+import { Button } from '../../lib';
```
rluders, line 4: Line 3 and 4 could be just one line.

rluders (changes requested):
I think that just one small change..

src/docs/pages/TimelinePage.tsx
joeychrys: > Line 3 and 4 could be just one line.

Done!

rluders:
@tulup-conner @bacali95 this one is a big one... May we have +1 review at least?

src/docs/pages/TimelinePage.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+const TimelinePage: FC = () => {
+  const examples: CodeExample[] = [
+    {
+      title: 'Default timeline',
+      code: (
+        <Timeline>
+          <Timeline.Item>
+            <Timeline.Point>
+              <div className="absolute -left-1.5 mt-1.5 h-3 w-3 rounded-full border border-white bg-gray-200 dark:border-gray-900 dark:bg-gray-700"></div>
```
bacali95, line 14: I'm seeing this getting repeated along the usage of the timeline component so I think if it can be moved within the Point component with adding some cutom logic to allow the use to pass an Icon as well.

src/lib/components/Timeline/Timeline.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+  component: Timeline,
+} as Meta;
+
+const Template: Story<TimelineProps> = (args) => <Timeline {...args} />;
+
+export const DefaultTimeline = Template.bind({});
+DefaultTimeline.storyName = 'Default Timeline';
+DefaultTimeline.args = {
+  children: (
+    <Timeline>
```
bacali95, line 17: The children here should be only the array of timeline items since this will be passed to the template in line 11.

src/lib/components/Timeline/TimelineItem.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export type TimelineItemProps = PropsWithChildren<ComponentProps<'li'>>;
+
+export const TimelineItem: FC<TimelineItemProps> = ({ children, ...props }) => {
+  const { horizontal } = useTimelineContext();
+  return (
+    <li
+      data-testid="timeline-item"
+      className={classNames({ 'mb-10 ml-6': !horizontal, 'relative mb-6 sm:mb-0': horizontal })}
+      {...props}
```
bacali95, line 13: For all components that we spread the rest of the props that the user can pass we need to extract the className prop and append it to the one we have otherwise it will get overridden.

bacali95 (changes requested):

src/lib/components/Timeline/TimelineItem.tsx
joeychrys: Oh I see what you mean because `...props` contains `className` as well. I'll fix it.

src/lib/components/Timeline/TimelinePoint.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export type TimelnePointProps = PropsWithChildren<{ icon?: FC<ComponentProps<'svg'>> }>;
+
+export const TimelinePoint: FC<TimelnePointProps> = ({ children, icon: Icon, ...props }) => {
+  const { horizontal } = useTimelineContext();
+  return (
+    <div data-testid="timeline-point" className={classNames({ 'flex items-center': horizontal })} {...props}>
+      {children}
+      {Icon ? (
+        <span className="absolute -left-3 flex h-6 w-6 items-center justify-center rounded-full bg-blue-200 ring-8 ring-white dark:bg-blue-900 dark:ring-gray-900">
```
tulup-conner, line 13: This needs to either not be visible to screen readers, or pass in an accessible name

src/lib/components/Timeline/TimelineTitle.tsx
```diff
@@ -0,0 +1,11 @@
+import { FC, PropsWithChildren, ComponentProps } from 'react';
+
+export type TimelineTitleProps = PropsWithChildren<ComponentProps<'h3'>>;
+
+export const TimelineTitle: FC<TimelineTitleProps> = ({ children, ...props }) => {
+  return (
+    <h3 className="text-lg font-semibold text-gray-900 dark:text-white" {...props}>
```
tulup-conner, line 7: This needs to be morphable into any heading level

tulup-conner (changes requested):

src/lib/components/Timeline/TimelineTitle.tsx
joeychrys: Should I make a prop called `tag` that is passed to the heading level. The user can then only pass in a range from 1-6 to the `tag` prop?

src/lib/components/Timeline/TimelinePoint.tsx
joeychrys: You might have to give me hand on this one. I'm not sure what the best way to do this is. :(

src/lib/components/Timeline/TimelineTitle.tsx
tulup-conner: Its a pretty common pattern to provide an as prop, like:

as="h5"

src/lib/components/Timeline/TimelinePoint.tsx
tulup-conner: I would probably just add the sr-only class (hide this icon for screen readers) since each icon will be accompanied by relevant text already.

joeychrys: > I would probably just add the sr-only class (hide this icon for screen readers) since each icon will be accompanied by relevant text already.

Okay so do I just add the sr-only class and make it a conditional based on whether or not there is an Icon. This would also replace the ternary statement right?

bacali95:
I think this PR is ready to go, do you have other requests @rluders @tulup-conner?

rluders (approved):

PR #119, opened 2022-05-15
set tab item button type to "button"

meddjelaili:
The tab items buttons should have a type of button. In case the tabs are used inside a form element, Clicking tabs won't submit the form.


rluders:
thanks, @meddjelaili

meddjelaili:
@rluders thanks for you too

PR #150, opened 2022-05-20
fix(component): Remove default export from `Breadcrumb`

tulup-conner:
`Breadcrumb` is uncallable right now in the standard convention:

```js
import {Breadcrumb} from "flowbite-react";
```

Since I `export default`'d. This changes that to a named export so the module can be imported like above.


rluders (approved):
Thank you.

PR #179, opened 2022-06-03
refactor(test): Migrate `jest` -> `vitest`

tulup-conner:
## Breaking changes

None.


tulup-conner:
Migrating to vitest was super easy, but it isn't entirely clear that it runs any faster when you run the whole test suite. It is much faster than jest when updating an individual test though.

tulup-conner:
It seems like the general consensus online is vitest ends up being about the same on average as jest on CI runs/full test runs, but is much faster in watch mode. I see a couple gains overall:

- No speed difference when running all of the tests
- Noticeably faster in watch mode/when running one or a few tests (i.e. while you are writing a test)
- Remove some create-react-app commands from package.json
- More accurate code coverage with c8 (our code coverage goes up by about 10% as well as a result)

But that's IMO. This is ready to merge whenever if we want to.

rluders (approved):
I tested it locally, didn't noticed anything wrong. It was a little bit faster than jest, true... but... I'm not sure if it is something that important.

However, I don't see any issues to get it merged. In fact, I liked the that we remove one piece of the dependency with CRA, but

rluders (approved):

PR #185, opened 2022-06-04
fix: `nanoid` requires Node `crypto` module

tulup-conner:
For users running Flowbite in a serverless environment like Cloudflare Workers, Node libraries like `crypto` are unavailable.

Fortunately, React added a hook that serves the exact purpose we need: [`useId()`](https://reactjs.org/blog/2022/03/29/react-v18.html#useid).

## Breaking changes

None.

## Bug fixes

- [x] [fix(component): Fix SSR mismatch in Tooltips](https://github.com/themesberg/flowbite-react/pull/185/commits/d4d95e7cca0ff9b32ae98e8fe92c566e8be65f49)
- [x] [fix(component): Replace nanoid -> useId()](https://github.com/themesberg/flowbite-react/pull/185/commits/31aafc9f4a54c091de0748055a03997b6084962c)

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,6 @@
     "@floating-ui/react-dom": "^0.6.3",
     "@floating-ui/react-dom-interactions": "^0.4.0",
     "classnames": "^2.3.1",
-    "nanoid": "^3.3.4",
     "react-icons": "^4.3.1",
     "react-indiana-drag-scroll": "^2.1.0"
   },
--- a/src/lib/components/Flowbite/index.tsx
+++ b/src/lib/components/Flowbite/index.tsx
@@ -1,5 +1,5 @@
 import type { FC, HTMLAttributes } from 'react';
-import { useLayoutEffect, useMemo } from 'react';
+import { useEffect, useMemo } from 'react';
 import { ThemeContext, useThemeMode } from './ThemeContext';
 import { mergeDeep } from '../../helpers/mergeDeep';
 import defaultTheme from '../../theme/default';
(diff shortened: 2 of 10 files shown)


src/lib/helpers/random-id.ts
```diff
@@ -0,0 +1,11 @@
+import windowExists from './window-exists';
+
+const randomId = (): string => {
+  if (windowExists()) {
+    return window.crypto.getRandomValues(new Uint32Array(1))[0].toString(16);
+  } else {
```
rluders, line 6: u don't need the `else` statement

rluders (changes requested):

src/lib/helpers/random-id.ts
tulup-conner: Yeah, I know that haha. I will change it if this style is preferred. I don't really think you gain much from a readability perspective, personally, which is why I don't do that.

tulup-conner:
My Cypress a11y workflow reports identical aria labels as a result of this since the SSR-rendered IDs are identical on a page with more than one instance of a component. I want to look into whether it's easy enough to disable those accessible labels until server rendered, so duplicate labels never get displayed.

tulup-conner:
The following additional small changes are necessary to support SSR, e.g. in Next or Remix:

- [fix(component): Fix SSR mismatch in Tooltips](https://github.com/themesberg/flowbite-react/pull/185/commits/13aab5287b914937118b573dd02adf624c301365)
- [refactor(component): Remove useLayoutEffect](https://github.com/themesberg/flowbite-react/pull/185/commits/aadb22c4e5cd6a4480f485a32af62d6fdf1303b1)

Pending this PR I can finally upgrade the starters to React 18.

tulup-conner:
This has a leftover file. Removing that.

rluders (approved):

PR #210, opened 2022-06-07
Fix: update typo on Footer.Link from Gihub to Github

Baxi19:
There is a typo in the Flowbite React documentation: 
**Gihub** appears but it should be **Github**

Screen Shot 2022-06-07 at 15 15 55


tulup-conner (approved):

rluders:
Thank you @Baxi19

PR #217, opened 2022-06-10
fix(component): Remove `type="button"` when not a `<button>`

tulup-conner:
`ListGroup.Item`s currently have `type="button"` always,
even though they can be an anchor (`<a>`) element
instead. That's changed now.

Also adds `props` since they weren't being passed
in for some reason.

## Breaking changes

None.

## Bug fixes

- [x] [fix(component): Remove type="button" when not a <button>](https://github.com/themesberg/flowbite-react/commit/36e3db8e9af86d5b646ac993865092d5887af247)


rluders:
@tulup-conner what is missing here? :)

tulup-conner:
As far as I can tell this is good to go, I'm not sure why I didn't ready it.

PR #226, opened 2022-06-13
fixed min-h to arbitrary values (rem)

grandehe4rt:
I have changed the css values from "common" px values to arbitrary values like "rem".

### DIFF
--- a/src/lib/theme/default.ts
+++ b/src/lib/theme/default.ts
@@ -240,7 +240,7 @@ export default {
     },
   },
   carousel: {
-    base: 'relative h-full w-full min-h-56 sm:min-h-64 xl:min-h-80 2xl:min-h-96',
+    base: 'relative h-full w-full min-h-[3.5rem] sm:min-h-[4rem] xl:min-h-[5rem] 2xl:min-h-[6rem]',
     indicators: {
       active: {
         off: 'bg-white/50 hover:bg-white dark:bg-gray-800/50 dark:hover:bg-gray-800',
--- a/src/lib/theme/default.ts
+++ b/src/lib/theme/default.ts
@@ -240,7 +240,7 @@ export default {
     },
   },
   carousel: {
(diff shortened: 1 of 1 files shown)


src/lib/theme/default.ts
```diff
@@ -240,7 +240,7 @@ export default {
     },
   },
   carousel: {
-    base: 'relative h-full w-full min-h-56 sm:min-h-64 xl:min-h-80 2xl:min-h-96',
+    base: 'relative h-full w-full min-h-[3.5rem] sm:min-h-[4rem] xl:min-h-[5rem] 2xl:min-h-[6rem]',
```
tulup-conner, line 243: Thank you! We should try to match with Flowbite itself and use the values provided by the original component here: https://flowbite.com/docs/components/carousel/

For example, the `2xl` breakpoint is set to `h-96` which translates to `24rem`

tulup-conner (changes requested):

grandehe4rt:
making clear that the carousels are now visible, i think my commit wasn't very clear.

src/lib/theme/default.ts
```diff
@@ -240,7 +240,7 @@ export default {
     },
   },
   carousel: {
-    base: 'relative h-full w-full min-h-56 sm:min-h-64 xl:min-h-80 2xl:min-h-96',
+    base: 'relative h-full w-full h-56 sm:h-64 xl:h-80 2xl:h-96',
```
rluders, line 243: I think that the we should replace the `h-full` and `w-full` with `h-min-full` and `w-min-full`, but we need to test. What do you think @tulup-conner?

rluders (changes requested):

src/lib/theme/default.ts
tulup-conner: I think that could be fine as well.

rluders:
Thanks @grandehe4rt

PR #237, opened 2022-06-15
Add className prop to button

dilan-dio4:
Previously, a custom className could be appended to the Button component. I noticed this functionality was removed in 0.0.26 and I implemented it to work with the button changes in 0.0.26

### DIFF
--- a/src/lib/components/Button/index.tsx
+++ b/src/lib/components/Button/index.tsx
@@ -78,7 +78,7 @@ const ButtonComponent: FC<ButtonProps> = ({
         outline && theme.outline.color[color],
         theme.base,
         theme.pill[pill ? 'on' : 'off'],
-        className
+        className,
       )}
       disabled={disabled}
       href={href}
--- a/src/lib/components/Button/index.tsx
+++ b/src/lib/components/Button/index.tsx
@@ -11,7 +11,7 @@ import { useTheme } from '../Flowbite/ThemeContext';
 import type { PositionInButtonGroup } from './ButtonGroup';
 import ButtonGroup from './ButtonGroup';
 
-export interface ButtonProps extends Omit<ComponentProps<'button'>, 'className' | 'color'> {
+export interface ButtonProps extends Omit<ComponentProps<'button'>, 'color'> {
   color?: keyof ButtonColors;
   gradientDuoTone?: keyof ButtonGradientDuoToneColors;
   gradientMonochrome?: keyof ButtonGradientColors;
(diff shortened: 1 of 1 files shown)


tulup-conner:
Hey there, this was a design choice, see #92

rluders:
@dilan-dio4 as @tulup-conner mentioned, it was a design choice. We are not allowing direct access to `className` props in order to keep the design system consistency. If there are any edge case, [it can be discussed here](https://github.com/themesberg/flowbite-react/discussions).

I'm closing the PR, since it will not be accepted for now.

dilan-dio4:
Hm. I understand that you want to unify the design process, but I guess I'm confused as what the benefit is of making components less customizable?

Overriding classnames directly makes it very easy to do one-off configurations, especially with tailwind integration.

rluders:
@dilan-dio4 you still can customize things by using the theme feature. The benefit of using the theme to customize the library components is that you keep the design consistency all over your application.

Even so... can you give me one example where the `className` is a *must have* and it cannot be replaced by the theme support?

dilan-dio4:
I appreciate the standard design system and I plan on using it. This package lives in the React + tailwind universe in which `className` is a standard of. What if I want to add a small margin to one button? Now I have to wrap in a Div? What is lost by keeping className as an option? Mui has a standard themeing system and also className.

It's not a deal breaker to me. This is fantastic library that I'll use either way. Just a small loss of functionality.

rluders:
I understand your point, @dilan-dio4. But we are not removing any TailwindCSS functionality. You still able to use TailwindCSS, as I said, you can customize it using the theme, and it will affect all the components, keeping your customization solid, and concise around all the places.

If you want to add some space between components, like the example that you gave, where you want to add a `margin` to the `button` you need to wrap it with a `div`. Try to see like this, the `margin` isn't a change to the button style, but to the layout/content structure of your system. So, this is why it does make sense to wrap it with a `div` and adding the `margin` to the `div`, not to the `button`. I don't know if I explain it correctly.

If you start to have access directly to the `className` prop to the design system component, you are going to have serveral miss-behaviors in your system.

1. Unpredictable styling behaviors (as @tulup-conner mentioned) - some components handle states and more.
2. You are going to start to have incontinence user UX
3. You are going to start to have code repetition (if you need to every time overwrite the class manually)
4. You are going to start to mix layout/content structure concepts to component concepts

Trying to clarify all the things:

- If you need that all your component have some property, use the theme support and add it to all;
- If you need one component to have a specific style, it is a custom component (outside the design system);
- If you have to add anything that doesn't actually change the button style, but it is related to the content structure, it doesn't belong to the component; in some cases can be done by adding something to the children or wrapping the component;

I do understand that other design system can support `className`, but as we said, this is a design choice, we are worried about UX, a11y, consistency, predictability and testability. This is why we are not allowing access directly to the component `className`.

And so far, I'm not convinced that given the access to it would bring any benefit that can't be achieved in some more predictable way.

PR #267, opened 2022-07-06
chore(docs): adds webpack to hugo

rluders:
n/a


rluders:
I could, but the flowbite original is using webpack, I basically just copied the things. Maybe we can change it in the future, but right now we need to put the documentation to work. :)

PR #311, opened 2022-07-26
React Component Preview to Hugo Documentation

rluders:
https://user-images.githubusercontent.com/191027/209187165-0f6553e2-6dc5-4f53-b141-d35954c3adde.mp4


rluders:
Maybe some helpful links:

https://codesandbox.io/s/l62ke?file=/src/App.js:511-527
https://reactjs.org/docs/add-react-to-a-website.html

docs/package.json
```diff
@@ -10,17 +10,22 @@
...
     "@babel/preset-env": "^7.18.6",
+    "@babel/preset-react": "^7.18.6",
     "@docsearch/js": "^3.1.1",
     "autoprefixer": "^10.4.7",
     "babel-loader": "^8.2.5",
     "core-js": "^3.23.3",
     "css-loader": "^6.7.1",
-    "css-minimizer-webpack-plugin": "^4.0.0",    
+    "css-minimizer-webpack-plugin": "^4.0.0",
+    "flowbite-react": "^0.3.4",
```
rluders, line 23: I'll later link it with the "current" build. For now, it just simplifies the test.

rluders:
Asking for an early review on this one. It isn't done yet, but I need some feedback. :)

tulup-conner:
@rluders Where are we at on this?

rluders:
> @rluders Where are we at on this?

This needs some work, but it is probably a good start. You should probably give it a try, and see if it works for you.

tulup-conner:
Eugh... I really don't like the workflow of this, having to go into a subdirectory and do an additional yarn install and run a separate server. I feel like we this is getting very complicated very fast. But I don't really have any solutions, so. It may be useful to bind these to `package.json` commands so there's less mental effort.

rluders:
Let's drop this one. It would be simpler to keep the React Documentation.

PR #315, opened 2022-07-31
fix(component): display text only on Sidebar.Item tooltip

tulup-conner:
## Breaking changes

None.

## Bug fixes

- [x] [fix(component): display text only on Sidebar.Item tooltip](https://github.com/themesberg/flowbite-react/commit/b44d784602eb6a78c4739d858c205848f226daa5)

## Tests

- [x] [test(component): drop false-positive Sidebar tests](https://github.com/themesberg/flowbite-react/commit/f5d4dae08a5a525995273f2a3c1d4161f6aa4590)

### DIFF
--- a/src/lib/components/Flowbite/FlowbiteTheme.ts
+++ b/src/lib/components/Flowbite/FlowbiteTheme.ts
@@ -413,7 +413,6 @@ export interface FlowbiteTheme {
       };
       content: {
         base: string;
-        collapsed: string;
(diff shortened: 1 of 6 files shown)


tulup-conner:
I've deliberately removed tests in this PR. These tests are unreliable. We should use cypress to test any hover interactions.

rluders (approved):
lgtm

PR #326, opened 2022-08-09
upgrade dependencies to latest

tulup-conner:
This involves migrating to the newest react testing library, which unfortunately is quite an overhaul.

## Breaking changes

None.

## Features

- [x] [feat(a11y): add role="dialog" to <Modal>s](https://github.com/themesberg/flowbite-react/pull/326/commits/47e219042d1c44d031817119daab34ac9d87b186)

## Bug fixes

- [x] [fix(a11y): add accessible label to Alert close/dismiss button](https://github.com/themesberg/flowbite-react/pull/326/commits/ab3c8083c1de3b809091bc132cc1a8c47c0d78f3)
- [x] [fix(a11y): add accessible label to <DarkThemeToggle> icons](https://github.com/themesberg/flowbite-react/pull/326/commits/b1bcd45920321f2ca5bc7403371342e76aff6868)
- [x] [fix(component): prevent <Accordion className="..">](https://github.com/themesberg/flowbite-react/pull/326/commits/e5c031d5ccb3cb12a476f3114dfea8ce07ffc6b4)

## Tests

- [x] [refactor(test): migrate tests to @testing-library/react 13.x](https://github.com/themesberg/flowbite-react/pull/326/commits/7885d928c89598ab4f4f598184550f039ed75fc2)


tulup-conner:
The coverage diff here isn't meaningful, nothing to worry about

rluders:
@tulup-conner I think that we should remove the codecov as a blocker to get the PR merged. What do you think?

tulup-conner:
Yea, probably in general a good idea. I think it is safe PRs will always have their coverage reviewed since Codecov makes it so easy with the message in the PR.

rluders:
@tulup-conner so, in this case, do you mind to include in this PR the configuration to the CI that make the codecov non-blocker? 

If I do remember correct is the `if_ci_failed` configuration that needs to be changed for it, but, better to check the docs: 
https://docs.codecov.com/docs/commit-status

rluders (approved):

codecov.yml
```diff
@@ -2,5 +2,6 @@ coverage:
   status:
     project:
       default:
+        if_ci_failed: success
         target: auto
         threshold: 0.5%
```
rluders, line 7: I guess that we are going to have to increase this one as well... maybe to 5%?

rluders (changes requested):

PR #342, opened 2022-08-24
Export TableContext

phbernard:
## Description

`TableContext` is now available for flowbite-react users.

As the context provides the `striped` and `hoverable` attributes, it can be useful for external components which want to fine tune the table based on one of these attributes.

Access to the context is also necessary to override one of the Table sub-components. For example, I want to write my own `TableComponent` because flowbite-react's has rounded corners which cannot be disabled.

## Type of change

## Breaking changes

none.

## How Has This Been Tested?

I was able to use `TableContext` in my own app.

## Checklist:

### DIFF
--- a/src/lib/components/Table/index.tsx
+++ b/src/lib/components/Table/index.tsx
@@ -37,4 +37,5 @@ export const Table = Object.assign(TableComponent, {
   Row: TableRow,
   Cell: TableCell,
   HeadCell: TableHeadCell,
+  Context: TableContext
 });
--- a/src/lib/components/Table/index.tsx
+++ b/src/lib/components/Table/index.tsx
@@ -37,5 +37,5 @@ export const Table = Object.assign(TableComponent, {
   Row: TableRow,
   Cell: TableCell,
   HeadCell: TableHeadCell,
-  Context: TableContext
+  Context: TableContext,
 });


src/lib/components/Table/index.tsx
```diff
@@ -37,4 +37,5 @@ export const Table = Object.assign(TableComponent, {
   Row: TableRow,
   Cell: TableCell,
   HeadCell: TableHeadCell,
+  Context: TableContext,
```
rluders, line 40: I think that it is better to expose the `useTableContext` helper, instead exposing the TableContext directly.

https://github.com/themesberg/flowbite-react/blob/e2b17a83c29af810d262fa4e7066ecdf40a56ac0/src/lib/components/Table/TableContext.tsx#L10

rluders (changes requested):

src/lib/components/Table/index.tsx
phbernard: That's right. Context consumers will prefer to use `useTableContext` instead of calling `useContext(TableContext)` manually. This is also much cleaner.

However, a public `useTableContext` doesn't address my personal use case. I want a table without rounded corners and shadow, which are both hard coded in the top level `div` of `TableComponent` in `src/lib/components/table/index.tsx`. This could be achieved by adding two `rounded` and `shadow` boolean attributes but I thought it was too specific. So the cleaner way I found to do this is to write my own `TableComponent` and use it in place of flowbite-react's original. Doing so explicitly requires to access `TableContext.Provider`.

Possible solutions;
- Export `useTableContext` in addition to `TableContext`. Users will rather use `useTableContext` instead of consuming the context directly while people like me will use `TableContext`.
- Only export `useTableContext`. In that case I need to find another solution to my problem, but that's not the topic of this PR.

rluders: In this case, if the issue is the style, I guess that we are going to the wrong direction here. I notice that the Table component doesn't yet support the Theme customization. Maybe, what you need can be achieved my introducing the theme support to the component, and not getting access to the TableContext.

phbernard: Okay. I'm rather new to the project, so if you feel like adding theme support is the best way to handle my case, you can just discard this PR. I already have a workaround so there is no emergency on my side.

Thank you for your help!

PR #434, opened 2022-11-16
feat(components): export components types

### LINKED ISSUE #433: Export all components interface/types to `flowbite-react`
Sometimes I need to handle some data from API or configurations that will later be passed to a flowbite-component. In this case, I would like to convert this data to the component property types.

All interfaces/types used by flowbite components must be exported as types to the default package.

None.

**Additional context**
One good example is when you are using the `react-router` to create the `Breadcrumb` and you need to handle some logic to the data to apply it to the `Breadcrumb. Item`. I can't import the `BreadcrumbItemProps` directly from the `flowbite-react` package.

Rafaelb4rros:
fix #433

## Description
list of components that the types were not being exported correctly:
image
Fixes #433

### DIFF
--- a/src/lib/components/Accordion/index.ts
+++ b/src/lib/components/Accordion/index.ts
@@ -1,1 +1,3 @@
 export * from './Accordion';
+export type { AccordionPanelProps } from './AccordionPanel';
+export type { AccordionTitleProps } from './AccordionTitle';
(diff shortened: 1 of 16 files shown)


rluders:
Thanks, @Rafaelb4rros

Issue #397, opened 2022-10-20
Add support for NextJS links like <Link href="/about"></Link>


andemohit:
**Is your feature request related to a problem? Please describe.**
I am using flowbite-react for the NextJS application. I am trying to use NextJS links inside the Sidebar component.

`<Sidebar.Item>
    <Link href="/about">About</Link>
</Sidebar.Item>`

In the console, it says `<a> cannot appear as a descendant of <a>`

**Describe the solution you'd like**
Could you please add support for `next/link`? It would be really helpful for my application.

PR #438, opened 2022-11-24
fix(sidebaritem): fix sidebar item with next link

### LINKED ISSUE #430: Maximum update depth exceeded with Next/Link and SidebarItem
**Describe the bug**
I opened a bug ticket here, see: https://github.com/vercel/next.js/issues/42902

**To Reproduce**
See: https://codesandbox.io/p/sandbox/hopeful-thompson-1lw449

**Expected behavior**
That it only renders once.

**Additional context**
This problem is occuring, because we're currently creating Components inside of other Components for e. g.: https://github.com/themesberg/flowbite-react/blob/v0.3.4/src/lib/components/Sidebar/SidebarItem.tsx#L34-L56

myasteiner:
## Description

This PR fixes the `next/link` component for the `SidebarItem`.

Fixes #430 

## Type of change

## Checklist:

### DIFF
--- a/src/lib/components/Sidebar/SidebarItem.tsx
+++ b/src/lib/components/Sidebar/SidebarItem.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import { ComponentProps, ElementType, FC, forwardRef, PropsWithChildren, useId } from 'react';
+import { ComponentProps, ElementType, FC, forwardRef, PropsWithChildren, ReactNode, useId } from 'react';
 import { Badge } from '../Badge';
 import type { FlowbiteColors } from '../Flowbite/FlowbiteTheme';
 import { useTheme } from '../Flowbite/ThemeContext';
@@ -21,6 +21,41 @@ export interface SidebarItemLabelColors extends Pick<FlowbiteColors, 'gray'> {
   [key: string]: string;
 }
 
+const ListItem: FC<PropsWithChildren<{ id: string; isCollapsed: boolean; tooltipChildren: ReactNode | undefined }>> = ({
+  id,
+  isCollapsed,
+  tooltipChildren,
+  children: wrapperChildren,
+}) => (
+  <li>
+    {isCollapsed ? (
+      <Tooltip content={<TooltipContent id={id}>{tooltipChildren}</TooltipContent>} placement="right">
+        {wrapperChildren}
+      </Tooltip>
+    ) : (
+      wrapperChildren
+    )}
+  </li>
+);
+
+const TooltipContent: FC<PropsWithChildren<{ id: string }>> = ({ id, children }) => (
+  <Children id={id}>{children}</Children>
+);
+
+const Children: FC<PropsWithChildren<{ id: string }>> = ({ id, children }) => {
+  const theme = useTheme().theme.sidebar.item;
+
+  return (
+    <span
+      className={classNames(theme.content.base)}
+      data-testid="flowbite-sidebar-item-content"
+      id={`flowbite-sidebar-item-${id}`}
+    >
+      {children}
+    </span>
+  );
+};
+
 const SidebarItem = forwardRef<Element, SidebarItemProps>(
   (
     { as: Component = 'a', children, icon: Icon, active: isActive, label, labelColor = 'info', className, ...props },
@@ -31,32 +66,8 @@ const SidebarItem = forwardRef<Element, SidebarItemProps>(
     const { isInsideCollapse } = useSidebarItemContext();
     const theme = useTheme().theme.sidebar.item;
 
-    const ListItem: FC<PropsWithChildren> = ({ children: wrapperChildren }) => (
-      <li>
-        {isCollapsed ? (
-          <Tooltip content={<TooltipContent>{children}</TooltipContent>} placement="right">
-            {wrapperChildren}
-          </Tooltip>
-        ) : (
-          wrapperChildren
-        )}
-      </li>
-    );
-
-    const TooltipContent: FC<PropsWithChildren> = ({ children }) => <Children>{children}</Children>;
-
-    const Children: FC<PropsWithChildren> = ({ children }) => (
-      <span
-        className={classNames(theme.content.base)}
-        data-testid="flowbite-sidebar-item-content"
-        id={`flowbite-sidebar-item-${id}`}
-      >
-        {children}
-      </span>
-    );
-
     return (
-      <ListItem>
+      <ListItem id={id} isCollapsed={isCollapsed} tooltipChildren={children}>
         <Component
           aria-labelledby={`flowbite-sidebar-item-${id}`}
           className={classNames(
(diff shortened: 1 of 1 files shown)


rluders:
I'm not sure about this one... maybe @tulup-conner could give it a look.

tulup-conner:
I don't totally understand what this does TBH. I don't really use Next. I don't think it makes sense for the library to have Next-specific patches.

rluders:
Yes. I agree with you one this one @tulup-conner, I don't want to see specific-nextjs or any specific stuff landing at the flowbite-react. I have the same feeling that this is too specific. I'm wondering if #430 can't be solved in some more generic approach.

rluders:
Also, I'm seeing the "Legacy" word there... So, is this for new Next.js version or just to support oldest versions? I mean, if this is a legacy support, I'll say: we don't care.

src/lib/components/Footer/FooterLink.tsx
```diff
@@ -1,19 +1,21 @@
...
 
-export const FooterLink: FC<FooterLinkProps> = ({ children, className, href }) => {
+export const FooterLink = forwardRef<HTMLAnchorElement, FooterLinkProps>(({ children, className, href }, ref) => {
   const theme = useTheme().theme.footer.groupLink.link;
 
   return (
     <li className={classNames(theme.base, className)}>
-      <a href={href} className={theme.href}>
+      <a href={href} className={theme.href} ref={ref}>
         {children}
```
rluders, line 15: BTW, I'm fine to have the `ref=` here. But I'm not sure about the remaining Next.js specific stuff.

src/lib/components/Navbar/NavbarLink.tsx
```diff
@@ -8,26 +8,31 @@ export interface NavbarLinkProps extends PropsWithChildren<ComponentProps<'a'>>
...
+          className={classNames(
+            theme.base,
+            {
+              [theme.active.on]: active,
+              [theme.active.off]: !active && !disabled,
+            },
+            theme.disabled[disabled ? 'on' : 'off'],
+            className,
+          )}
+          ref={ref}
```
rluders, line 28: Same stuff here.

src/lib/components/Next/NextLinkLegacyBehavior.tsx
```diff
@@ -0,0 +1,15 @@
+import { useCallback, useState, type FC, type ReactElement } from 'react';
```
rluders, line 1: This component seems to be like a "documented solution" not a flowbite-react component. So, maybe we could add it to the documentation insted? Some "Known Issues" section?

src/lib/components/Sidebar/SidebarItem.tsx
```diff
@@ -31,32 +88,8 @@ const SidebarItem = forwardRef<Element, SidebarItemProps>(
...
-        data-testid="flowbite-sidebar-item-content"
-        id={`flowbite-sidebar-item-${id}`}
-      >
-        {children}
-      </span>
-    );
-
     return (
-      <ListItem>
+      <ListItem id={id} isCollapsed={isCollapsed} contentChildren={children}>
```
rluders, line 92: I felt like having the same `id` there and at line 144 will be a problem. I mean, the `id` should be always unique. Right? Maybe adding some suffix to the one at 144 line, something like `${id}-children`.

rluders (changes requested):

esko22:
Been following this a bit as I am having the same issue reported in #397 using the vanilla  { Link } from "react-router-dom" producing the `<a> cannot appear as a descendant of <a>.` using 0.3.5. 

Some of this conversation is pushing my boundaries of component rendering :) but just wanted to add a note that the issue doesn't appear to be isolated to the NextJs Link component. Not sure if there are any nuances with how NextJs extends from the core React lib or if it even does at all but I've noticed that the NextJs link shows to use `href` in the docs where the react router specifies `to`

```
        <Sidebar.Items className="pt-8 pr-12">
          <Sidebar.ItemGroup>
                  <Sidebar.Item className="mb-2" icon={IoMdPin}>
                      <Link to="/" >
                      Map
                      </Link>
                  </Sidebar.Item>
            </Sidebar.ItemGroup>
        </Sidebar.Items>
```

tulup-conner:
Well, I've brought this up somewhere on this repo before, but the universal standard for component libraries as far as I understand is to provide a prop, generally called `as`, that lets you essentially "cast" any flowbite-react elements that are/can be links to any custom link component. 

```js
<Sidebar.Item className="mb-2" icon={IoMdPin} as={Link} to="/">Map</Sidebar.Item>
```

I tried to do this a few months ago but I'm too stupid. Or, perhaps I just need to try again. Anyway, I think this is probably the most familiar and ergonomic solution.

rluders:
I'm just wondering what would happen if the `Sitebar.Item` became a simple `div` if there is no `href` property attached to it, and if there is a `href` it should become an `a` otherwise, it will use the `children`. So, in this case, could be a `Link` or a `Button` or whatever fits to the desired action.

And all that I'm saying right now could be really dumb... I'm just wondering what could happen. 🤔

tulup-conner:
That makes sense to me, too. I have a theory if we could figure out how to do the `as={}` it might be less code/more readable, but like I said I failed before.  I'll look into this.

rluders:
I'm not sure if I would take my suggestion seriously. I have no idea how it is going to work and if it is a good practice at all. It was just a crazy idea. 🤣

I think that also using the `as={}` could be interesting for `Button` component and so many other components that may need nome "tag mutation".

tulup-conner:
Ok, I tried again and the issue I was having with Sidebar.Item is that you can't override className, which is how Remix facilitates self-aware links via NavLink. That's because we designate it ourselves in Sidebar.Item on L62. It looks like all other properties work just fine.

I'm not sure how to work around this - it might be that I will just need to patch Remix on my apps. I'm not sure if it is something that will come up with other users/use cases where they need to override className for the as={} component and can't.

tulup-conner:
> Ok, I tried again and the issue I was having with Sidebar.Item is that you can't override className, which is how Remix facilitates self-aware links via NavLink. That's because we designate it ourselves in Sidebar.Item on L62. It looks like all other properties work just fine.
> 
> I'm not sure how to work around this - it might be that I will just need to patch Remix on my apps. I'm not sure if it is something that will come up with other users/use cases where they need to override className for the as={} component and can't.

The workaround looks like...

Patch to `flowbite-react`

```js
-          className={classNames(
-            theme.base,
-            isActive && theme.active,
-            !isCollapsed && isInsideCollapse && theme.collapsed.insideCollapse,
-            className,
-          )}
+          className={
+            typeof className === 'string' || typeof className === 'undefined'
+              ? classNames(
+                  theme.base,
+                  isActive && theme.active,
+                  !isCollapsed && isInsideCollapse && theme.collapsed.insideCollapse,
+                  className,
+                )
+              : className
+          }
```

Custom component for ease of use

```js
export const SidebarNavLink: FC<NavLinkProps & SidebarItemProps> = function ({
  children,
  ...props
}) {
  const theme = useTheme().theme.sidebar.item;

  return (
    <FlowbiteSidebar.Item
      as={NavLink}
      className={({ isActive }: { isActive: boolean }) =>
        classNames(theme.base, isActive && theme.active)
      }
      {...props}
    >
      {children}
    </FlowbiteSidebar.Item>
  );
};
```

Actual links

```js
<Sidebar.NavLink to="/brand/components/tooltip">Tooltip</Sidebar.NavLink>
```

The end product in the final snippet here is an ideal DX from my perspective, but I think the pieces in between can be better.

rluders:
@myasteiner Well, I guess that this now changes the PR general goal. So, maybe we could update the title and the description.

rluders (approved):

PR #439, opened 2022-11-25
feat(avatar): adds color prop and test case for it

### LINKED ISSUE #409: Support colors in Avatar component
Avatar component has color property, but it doesn't work. Any color that is passed doesn't change anything.
IMO, it should have the following colours:
- dark
- failure
- gray/info/light
- purple
- success
- warning
- pink

Gray, info and light, if all added, should have different tones, to distinguish them.

I've tried to implemented and almost got it to work, so I think whoever picks this should:
- Change `default.ts` and add color styling on 'avatar' key value (see snippets below)
- `Avatar.tsx` must extends from Omit instead of PropsWithChildren and have 'color' argument passed. It should have another prop called `color` with a custom interface that explicit tells the colour options. Add `theme.color[color]` at least on the placeholder HTML element.
- Add the custom interface from previous step into `FlowbiteTheme.ts` on avatar key.

sldk-yuri:
1. Adds color property to Avatar component to change image ring color. Works only if border property is true. 
2. Adds 'color' test case to the Avatar`s Theme test case

#409

## Description

Implements avatar component enhancement introduced by @tiagossa1 The component has ability to change image ring color via prop `color` if `bordered` prop is true.

Fixes #409

## Type of change

## How Has This Been Tested?

`'should use custom colors'` case was added to `'Components / Avatar'` -> `'Theme'` in Avatar.spec.tsx 
Storybook story works correct without any updates.

## Checklist:

### DIFF
--- a/src/docs/pages/AvatarPage.tsx
+++ b/src/docs/pages/AvatarPage.tsx
@@ -24,6 +24,52 @@ const AvatarPage: FC = () => {
       ),
     },
     {
+      title: 'Colored Avatar',
+      code: (
+        <>
+          <div className="flex flex-wrap gap-2">
+            <Avatar img="https://flowbite.com/docs/images/people/profile-picture-5.jpg" rounded bordered color="gray" />
+            <Avatar
+              img="https://flowbite.com/docs/images/people/profile-picture-5.jpg"
+              rounded
+              bordered
+              color="light"
+            />
+            <Avatar
+              img="https://flowbite.com/docs/images/people/profile-picture-5.jpg"
+              rounded
+              bordered
+              color="purple"
+            />
+            <Avatar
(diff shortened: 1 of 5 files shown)


rluders:
@sldk-yuri is it good to review? :)

sldk-yuri:
@rluders yes, it's. Take an additional look at default color pallete for avatar component `lib/default.ts` from line 70 to 80, please. Probably we may change/update it with new colors. Thank you.

rluders:
I think this is fine as it is... AFAIK, people still can extend the colors as their like, and add some custom colors to their own theme. Thank you, I'm accepting this PR.

sldk-yuri:
@rluders Thank you.

PR #483, opened 2022-12-09
feat(Avatar): Support overriding the Avatar img element. Fixes #482

### LINKED ISSUE #482: Allow the <img> component in <Avatar> to be overridden by the user
When creating a website, I used Google SSO for logging in the user. When displaying the profile image (which is on Google's domain), I had no way to override the <Avatar> image to add `referrerPolicy="no-referrer"`. I'm sure there are other situations when having direct access to the underlying element would be helpful as well.

I also would like to use Next.js's <Image> component with <Avatar>, but that's not possible without forking the component entirely.

Add a new prop to <Avatar> called `imgElement`. This would be a render prop which lets the user render whatever component they want in place of . This makes it possible to add extra props to the , or replace it with Next.js < Image > or any other component. 

The existing props passed to the  component in <Avatar> would be passed to the render prop function so they can be spread on the user provided component.

```jsx
<Avatar src={avatarUrl} rounded imgElement={(props) => <img {...props} referrerPolicy="no-referrer" />} />
<Avatar src={avatarUrl} rounded imgElement={(props) => <Image {...props} width={64} height={64} />} />
```

The render prop solution is elegant because it solves for both adding props to the  and changing the underlying element without adding a ton of extra props. This would be a non-breaking change, because omitting the optional `imgElement` prop would fall back on the existing behavior.

My workaround for this was to fork the <Avatar> element entirely by copying its source code into my project. 

**Additional context**
If this issue is accepted, I intend to implement it and submit a PR myself.

alexanderson1993:
## Description

Changes the Avatar `img` prop to either be a string or a render prop that returns a React element. That makes it so any of the props can be overridden, or an entirely different element could be used in its place.

The docs site doesn't support the render prop syntax by default, so I added a small extension that allows the docs pages to provide a `rawCode` property which is passed directly to the syntax highlighter instead of generating the code sample from the `code` JSX.

Fixes #482

## Type of change

## How Has This Been Tested?

- [x] Automated test that renders using the new render props and checks to see if a special attribute is included on the rendered  element.

## Checklist:

### DIFF
--- a/src/docs/pages/AvatarPage.tsx
+++ b/src/docs/pages/AvatarPage.tsx
@@ -61,6 +61,66 @@ const AvatarPage: FC = () => {
       ),
     },
     {
+      title: 'Override Image Element',
+      code: (
+        <div className="flex flex-wrap gap-2">
+          <Avatar
+            img={(props) => (
+              <img
+                referrerPolicy="no-referrer"
+                src="https://flowbite.com/docs/images/people/profile-picture-5.jpg"
+                {...props}
+              />
+            )}
+          />
+          <Avatar
+            img={(props) => (
+              <picture>
+                <source
+                  media="(min-width: 900px)"
+                  srcSet="https://flowbite.com/docs/images/people/profile-picture-3.jpg"
+                />
+                <source
+                  media="(min-width: 480px)"
+                  srcSet="https://flowbite.com/docs/images/people/profile-picture-4.jpg"
+                />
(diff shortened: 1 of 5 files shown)


tulup-conner:
LGTM, I think the logic in this component is getting a bit complicated to read but I don't know how to fix that at all.

tulup-conner (approved):

PR #486, opened 2022-12-12
Expose setActiveTab to ref

bestan:
## Description

I have a use case where I'd like to switch to Tab 2 after a performed action on Tab 1, i.e. I'd either need `setActiveTab` exposed or props enabling the switch.

Playing around with `TabItemProps.active`, I got to a first iteration where `TabsComponent` has an extra `useEffect` that switches active tab reacting to `TabItemProps.active` changes. However, this approach presented one major bug, consider the following sequence of action and effects:
1. Tab 2 has `TabItemProps.active == true`
2. User switches to Tab 3 (using standard top buttons)
3. Parent component attempts to switch back to Tab 2, but Tab 2 still has `TabItemProps.active == true`, so `useEffect` is not called.

Syncing state and props to resolve that felt too complex and invasive.

Instead, I forwarded a ref that exposes `setActiveTab` back to the parent.

Let me know if you support such approach and I can update documentation.

## Type of change

## How Has This Been Tested?

- [x] Test A
```
import { useRef } from 'react';
import { Button, Tabs, TabsRef } from '../../lib';

const TabsPage = () => {
  const ref = useRef<TabsRef>(null);

  return (
    <>
      <Tabs.Group aria-label="Default tabs" style="default" ref={ref}>
        <Tabs.Item title="Profile">
          Profile content
        </Tabs.Item>
        <Tabs.Item title="Dashboard">Dashboard content</Tabs.Item>
        <Tabs.Item title="Settings">Settings content</Tabs.Item>
        <Tabs.Item title="Contacts">Contacts content</Tabs.Item>
        <Tabs.Item disabled title="Disabled">
          Disabled content
        </Tabs.Item>
      </Tabs.Group>
      <Button onClick={() => ref.current?.setActiveTab(0)}>Tab 0</Button>
      <Button onClick={() => ref.current?.setActiveTab(1)}>Tab 1</Button>
      <Button onClick={() => ref.current?.setActiveTab(2)}>Tab 2</Button>
    </>
  );
}

export default TabsPage;
```

## Checklist:

### DIFF
--- a/src/docs/pages/TabsPage.tsx
+++ b/src/docs/pages/TabsPage.tsx
@@ -1,11 +1,14 @@
-import type { FC } from 'react';
+import { FC, useRef, useState } from 'react';
 import { HiAdjustments, HiClipboardList, HiUserCircle } from 'react-icons/hi';
 import { MdDashboard } from 'react-icons/md';
-import { Tabs } from '../../lib';
+import { Button, Tabs, TabsRef } from '../../lib';
 import type { CodeExample } from './DemoPage';
 import { DemoPage } from './DemoPage';
 
 const TabsPage: FC = () => {
+  const [activeTab, setActiveTab] = useState<number>(0);
+  const tabsRef = useRef<TabsRef>(null);
+
   const examples: CodeExample[] = [
     {
       title: 'Default tabs',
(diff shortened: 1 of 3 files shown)


bestan:
Prettier adjusted indentation, hiding whitespaces can be very helpful when reviewing this PR on GitHub.

bestan:
It would also make sense to expose activeTab. Let me know what you think.

rluders (commented):
Thanks, @bestan.

I think that it makes sense. I guess that, if @tulup-conner didn't see anything wrong here we can accept it. Just, please, do not forget to also add to the PR:

- [ ] ~Storybook~
- [x] Documentation
- [x] Unit test

bestan:
Added docs and tests. Not sure what to add to the storybook since it isn't a visual change.

I decided against exposing activeTab in the ref because changing tabs does not re-render parent component, meaning that ref.activeTab would be out-of-date most of the time.

Instead, I added onActiveTabChange callback to props.

Docs have an example of using both setActiveTab and onActiveTabChange.

rluders:
Thank you, @bestan!

PR #33, opened 2022-04-18
feat: add ability to drag to scroll for the Carousel component

bacali95:

PR #552, opened 2023-01-19
Fixes Incorrect change of Carousel indicators when dragging #531

### LINKED ISSUE #531: incorrect change of Carousel indicators when dragging from 1 to 0
**Describe the bug**

Carousel indicator doesn't change its state when dragging from 1 to 0, it just keep staying at 1.

https://user-images.githubusercontent.com/91225197/211947732-43730e6c-1468-44b4-91df-baeb20dc4ec7.mp4

________________

**To Reproduce**
1. Go to https://flowbite-react.com/carousel
2. Try to drag from 1 to 0 slide 
3. See error

_________________

Cause of such behavior is the lastest part of condition down below:

```jsx

  useEffect(() => {
    if (carouselContainer.current && !isDragging && carouselContainer.current.scrollLeft !== 0) {
      setActiveItem(Math.round(carouselContainer.current.scrollLeft / carouselContainer.current.clientWidth));
    }
  }, [isDragging]);

```

Without that condition Carousel works well (see exp. behavior). When dragging from 1 to 0 `carouselContainer.current.scrollLeft` should be 0 to properly calculate a new active item id.

___________________

**Been trying to commit a fix** but 6/9 of Carousel tests had failed.

_1. Carousel > should render and show first item
2. Carousel > Controls > should change slide via click on indicator
3. Carousel > Controls > should change slide via click on control
5. Carousel > Items traversal > should go through items by default
6. Carousel > Items traversal > should go through items after given seconds
7. Carousel > Items traversal > should not to go through items_

I have checked each case manually and the component works properly.

**Could there be any problem with the tests?
I've found an issue https://github.com/themesberg/flowbite-react/issues/49 where that condition was added. Lines 56-57 [here](https://github.com/themesberg/flowbite-react/commit/5c9c9fad075a7279a8bc1e6d25afe66ec0acbd22?diff=split#diff-2c62e3546e9ff3749bbc8501b979e72681bf6b108e1fee790f0889bbfd5038c8).**

___________________

**Expected behavior**

https://user-images.githubusercontent.com/91225197/211943071-399c53b2-1fcd-439f-897a-571118b32281.mp4

_________________

**System information:**
OS: Windows 10
Browser: Chrome, also tested in Edge and Safari (iOS)

CorwinDev:
Fixes the issue by remove !== 0, it is 0 if the first slide is selected

Fixes #531

-  Bug fix (non-breaking change which fixes an issue)


rluders:
@CorwinDev could you please fix your commit message to use [Conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)?

tulup-conner:
@CorwinDev Could you state more clearly (possibly example?) what this is doing?

CorwinDev:
> @CorwinDev Could you state more clearly (possibly example?) what this is doing?

It remove's the check !== 0 as the first slider is always 0

tulup-conner:
When I try to use the Carousel on mobile and swipe between slides, it just doesn't scroll predictably at all. For me it scrolls like a normal box, and I can even scroll vertically for some reason. I would expect it to go to the next slide in a single animation when I swipe to the left.

@CorwinDev I'm still not entirely sure I understand what your specific problem is. That's partially because I can't use Carousel on mobile by swiping in a predictable way in general. Like, I have the same problems even if I start at a slide other than the 1st slide.

I think this component has a few fundamental issues we will want to look at between 0.4.0 and 1.0.0.

tulup-conner (approved):

rluders:
@CorwinDev unit test need to be fixed.

PatrickCleary:
> I can't use Carousel on mobile by swiping in a predictable way in general. Like, I have the same problems even if I start at a slide other than the 1st slide.

@tulup-conner 

I have the same problems. Given that the standard flowbite does not have dragging capability, and it seems to have many issues I would propose to just remove the functionality from the component entirely. Would likely just need to revert this PR: https://github.com/themesberg/flowbite-react/pull/33/files#diff-eb39cfedc1aa12f1be9bda3c5be9f138d7289fae71c463d8255b5cab92e734d4

Issue #473, opened 2022-12-06
Support for all theme colors in components (pink, teal, etc.)?


rluders:
All components that support the property `color=` MUST support all `FlowbiteColors`. 

Currently, we are using `Pick` to select the ones that we want. Not ideal, I would say/. It would be better if we support all the colors.

It doesn't mean that we should/must provide styling at the theme for all these colors, just support removing the `Pick` probably is good enough.  (NEED TO BE TESTED - I do accept other solutions ideas)

### Discussed in https://github.com/themesberg/flowbite-react/discussions/325

<div type='discussions-op-text'>

<sup>Originally posted by **dasveloper** August  9, 2022</sup>
My application is based around the "pink" color from Tailwind's default theme colors. However with flowbite-react, the Button, Progress, Dropdown, etc. components don't seem to support `color="pink"`. Radio and Checkbox don't appear to support colors at all?

Regular Flowbite [support pink](https://flowbite.com/docs/customize/colors/) and some components like the badge do too. Just wondering if that is something that can be updated or if I'm missing something?</div>

lirbre:
Hey @rluders. I'm looking foward to start contributing. 

Can you provide me some context? This issue had some progress? This seems to be a pretty good issue to start collaborating. :)

PR #586, opened 2023-02-09
feat(/lib/theme): add Tailwind CSS colors to  `<Alert>`s

lirbre:
## Description

_(It's my first PR here lol. Looking forward to fix things if needed. :))_ 

This change starts adding all colors to components in Flowbite. I'll be partially updating the components until I got every of then with all colors from FlowbiteColors avaliable. 

Fixes # [473](https://github.com/themesberg/flowbite-react/issues/473) _(partially)_

## Type of change
- [X] New feature

## How Has This Been Tested?
Since it was a visual change I used the Storybook to validate if colors are changing as expected.

- [x] Storybook

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] Any dependent changes have been merged and published in downstream modules


src/lib/components/Alert/Alert.tsx
```diff
@@ -26,7 +26,25 @@ export interface FlowbiteAlertCloseButtonTheme {
   color: AlertColors;
 }
 
-export interface AlertColors extends Pick<FlowbiteColors, 'failure' | 'gray' | 'info' | 'success' | 'warning'> {
+export interface AlertColors
+  extends Pick<
```
rluders, line 30: You don't need to use `Pick<>` here if you are getting all the colors from the `FlowbiteColors` interface, you just need to `AlertColors extends FlowbiteColors`. You could even, or probably remove the `AlertColors` interface, and use the `FlowbiteColors` in its place.

rluders (changes requested):

src/lib/components/Alert/Alert.tsx
lirbre: Surely!

lirbre: The problem here is that I didn't added "green", "yellow" and "red" because it would be redundant to "success", "error" and "warning".

I'll add them and implement that or just replace it with an Exclude.

lirbre: Did the "redundant" way so I could implement FlowbiteColors directly. Also added the Storybook argTypes to make it easier to test.

src/lib/components/Alert/Alert.tsx
```diff
@@ -43,7 +39,7 @@ export interface AlertProps extends PropsWithChildren<Omit<ComponentProps<'div'>
 export const Alert: FC<AlertProps> = ({
   additionalContent,
   children,
-  color = 'info',
+  color = 'pink',
```
tulup-conner, line 42: Can you revert the change to this line? We're good to go with this PR at that point

tulup-conner (changes requested):

src/lib/components/Alert/Alert.tsx
lirbre: done!

tulup-conner (approved):
LGTM

tulup-conner:
I can't merge this without a re-review from you @rluders

tulup-conner (approved):

rluders (approved):
lgtm. Thank you.

PR #608, opened 2023-02-17
fix(/lib/components/table): prevent scrollbars around `<Table>`s

### LINKED ISSUE #605: Scrollbars appear around tables when resizing the window or when table content is modified
**Describe the bug**
Quite often, tables are dynamic and rows are added, removed, hidden, shown, etc. Since the `<div>` that wraps the `<table>` has the class `overflow-x-auto`, this creates scrollbars any time the content of the table is modified. This also happens when resizing the window.

I suggest either removing `overflow-x-auto sm:rounded-lg` from the wrapper element, or making the rounded corners optional with a note in the docs that it is not recommended for tables that will be modified dynamically.

**To Reproduce**
Add any `<Table />` component then add or remove a row dynamically.

**Expected behavior**
No scrollbars should appear

**Project information:**

- Tailwind: 3.2.6
- Flowbite: 1.6.3
- Flowbite React: 0.3.8
- Type: Nextjs

multiwebinc:
## Description

This moves the background color and border radius to the cells and places a div under the table for the box shadow effect. That way if a table is modified (rows or columns added or removed) or if the screen resolution changes, it won't create scrollbars around the table.

Fixes #605

## Type of change

## How Has This Been Tested?

Just tested by trying it out

## Checklist:


tulup-conner:
I love this idea, but we don't want to ship custom CSS. Can you find a way to accomplish this in Tailwind CSS via the theme instead of creating a stylesheet file?

multiwebinc:
@tulup-conner What about like this?

tulup-conner:
We will want a solution using Tailwind CSS purely, which has support for custom selectors like these. Inline CSS has other negative implications for users, e.g., it will break Content Security Policies.

src/lib/components/Table/Table.tsx
```diff
@@ -34,10 +35,30 @@ const TableComponent: FC<TableProps> = ({ children, striped, hoverable, classNam
   return (
     <div data-testid="table-element" className={classNames(theme.wrapper)}>
       <TableContext.Provider value={{ striped, hoverable }}>
+        <div className={classNames(theme.shadow, className)}></div>
         <table className={classNames(theme.base, className)} {...props}>
           {children}
         </table>
       </TableContext.Provider>
+      <style jsx>{`
```
rluders, line 43: I don't think that we want to do this. You must use TailwindCSS classes. Flowbite is a TailwindCSS-based component library, there is no point to use CSS or SASS or whatever.

rluders (changes requested):

src/lib/components/Table/Table.tsx
rluders: https://tailwindcss.com/docs/utility-first

multiwebinc:
@rluders @tulup-conner I was looking into the [`group-{modifier}` class](https://tailwindcss.com/docs/hover-focus-and-other-states#styling-based-on-parent-state), however I need 2 levels of groups, because it needs to select the first/last child of `<table>` combined with the first/last `<td>` or `<th>`. Would you be opposed to using https://www.npmjs.com/package/tailwindcss-scoped-groups? 

If not, the only thing I can do with tailwind is style the top borders in `<thead>` (if there is one) and bottom borders of `<tfoot>` (if there is one). Without knowing whether `<tbody>` is the first or last child, you can't add a border radius to any of the cells in it. Unless maybe I add two more props to the `<Table>` component - `headless` and `footless` and if either of these are `true`, it will add the border radius to the corresponding tbody cells.

multiwebinc:
@rluders @tulup-conner Ok, I think that's it. No stylesheets. I didn't know about nested groups, so I didn't think it was possible without stylesheets.

rluders:
@multiwebinc could you sync your branch with the `main` branch? It has some conflicts because the theme structure was changed. So, you may have to adjust a few things on your PR.

multiwebinc:
@rluders What do you think, would it make sense to move the `table.cell` interface to `table.body.cell` to avoid confusion with `table.head.cell`?

rluders:
> @rluders What do you think, would it make sense to move the `table.cell` interface to `table.body.cell` to avoid confusion with `table.head.cell`?

I think that it makes sense. WDYT, @tulup-conner ?

rluders:
Well, I can't see any reason to keep this one pending. I'm accepting it, @multiwebinc. Thank you!

PR #650, opened 2023-03-04
fix(themecontext): fixing `ReferenceError: window is not defined` error with Next.js

fsioni:
## Description

The variable `userPreferenceIsDark` was initialized with a value from the window object. This was causing an error with Next because of SSR. I added the check of the window, if it's on the server side the value is set to `false`

## Type of change

## How Has This Been Tested?

On my Next.js project that was causing the error.

## Checklist:

### DIFF
--- a/src/lib/components/Button/Button.tsx
+++ b/src/lib/components/Button/Button.tsx
@@ -64,6 +64,7 @@ export interface ButtonProps extends Omit<ComponentProps<'button'>, 'color' | 'r
   gradientDuoTone?: keyof ButtonGradientDuoToneColors;
   gradientMonochrome?: keyof ButtonGradientColors;
   href?: string;
+  target?: string;
   label?: ReactNode;
   outline?: boolean;
   pill?: boolean;


src/lib/components/Flowbite/ThemeContext.tsx
```diff
@@ -31,7 +31,8 @@ export function useTheme(): ThemeContextProps {
 }
 
 export const useThemeMode = (): [Mode, React.Dispatch<React.SetStateAction<Mode>>, () => void] => {
-  const userPreferenceIsDark = () => window.matchMedia?.('(prefers-color-scheme: dark)').matches;
+  const userPreferenceIsDark = () =>
+    windowExists() ? window.matchMedia?.('(prefers-color-scheme: dark)').matches : false;
```
tulup-conner, line 35: windowExists() && `...` instead of the ternary and good to go

src/lib/components/Footer/FooterCopyright.tsx
```diff
@@ -31,7 +33,7 @@ export const FooterCopyright: FC<CopyrightProps> = ({
     <div data-testid="flowbite-footer-copyright" className={classNames(theme.base, className)} {...props}>
       © {year}
       {href ? (
-        <a href={href} className={theme.href}>
+        <a href={href} target={target} className={theme.href}>
```
tulup-conner, line 36: reset your branch, this is an unrelated file

tulup-conner (changes requested):

PR #601, opened 2023-02-16
Prevent modal content from being truncated

multiwebinc:
Fixes #600

## Description

Prevent modal content from being truncated

## Type of change

Please delete options that are not relevant.

- Bug fix (non-breaking change which fixes an issue)

## Breaking changes

Possibly. I'm not sure.

## How Has This Been Tested?

See issue.

## Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes
- [ ] Any dependent changes have been merged and published in downstream modules
- [x] I have checked my code and corrected any misspellings

Issue #600, opened 2023-02-16
Modal content is truncated at the top on medium displays with long content


multiwebinc:
**Describe the bug**
Modal content is truncated at the top on medium displays with long content

**To Reproduce**

I have the following modal:

```jsx
<Modal show={true} onClose={onClose}>
  <Modal.Header>Header</Modal.Header>
  <Modal.Body>
    <ol className='list-decimal'>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
      <li>List item</li>
    </ol>
  </Modal.Body>
  <Modal.Footer>
    <Button color='gray' onClick={onClose}>
      Cancel
    </Button>
    <Button onClick={onSave}>Save</Button>
  </Modal.Footer>
</Modal>
```

On small screens, it displays fine, however on larger screens (`md` or greater), the top of the modal is truncated.

**Expected behavior**

**Screenshots**
Small screen:

![image](https://user-images.githubusercontent.com/901732/219267173-5e77917d-9eaf-42bd-b71d-650c57bc542f.png)

Medium screen or greater (note, this is scrolled all the way to the top):

![image](https://user-images.githubusercontent.com/901732/219267240-7114c081-375f-43f0-8f20-d88723da5234.png)

**Project information:**

- Tailwind: 3.2.6
- Flowbite: 1.6.3
- Flowbite React: 0.3.8
- Type: Nextjs

alex-griffiths:
[...]

Issue #537, opened 2023-01-14
Modal outflows on top


peteracs:
**Describe the bug**
If the default modal has too much contents and the screen width is >=768 px, the modal outflows on the top, hiding the contents.

**To Reproduce**
Create a default modal, fill it up with a lot of content that should stretch the height of it. It will be okay with screen-width <768, not working >768px

**Screenshots**
![Screenshot from 2023-01-14 00-39-31](https://user-images.githubusercontent.com/5214842/212457801-c6cad8b0-6445-43d5-9da0-6b69e168a0c4.png)

![Screenshot from 2023-01-14 00-39-36](https://user-images.githubusercontent.com/5214842/212457802-9c085bab-c299-4ed0-a5da-5890b657ee38.png)

Seems like the items-center class on the modal cause the issue. If I set the modal to align-items: normal; the issue is gone.

rluders:
Maybe it is related to #520 somehow?

Allam0053:
I used this in my code
```
      <Modal
        className='!h-full !items-start'
      >
         // children
      </Modal>
```

[![modal overflow](https://img.youtube.com/vi/3aMeQKNhGXU/0.jpg)](https://www.youtube.com/watch?v=3aMeQKNhGXU)

basically, it overwrites the base styling 'items-center' and the '!h-full' just because I want to ensure that the height fills the screen.

this is only for overflowing elements, I didn't test it in non-overflowing elements

Issue #520, opened 2023-01-06
Modal backdrop does not cover full screen when viewport size is small


zammitjohn:
**Describe the bug**
Modal backdrop does not cover full screen when viewport size is small

**To Reproduce**

1. Open https://flowbite-react.com/modal
2. Toggle modal
3. Resize window size width to < 767 px

**Expected behaviour**
Backdrop should always cover full screen

**Screenshots**
![Screenshot 2023-01-06 124911](https://user-images.githubusercontent.com/64637336/211006858-063a6f50-c1e8-41b7-803f-e363ab83916b.png)

mortezasabihi:
The bug is caused by the **core flowbite** found at this link: https://github.com/themesberg/flowbite/blob/e8c6490bd1893f467f347c41dc5147e6648b9dcb/plugin.js#L455.
The purpose of this line modal: 'calc(100% - 2rem)' is unclear, could you provide some context @zoltanszogyenyi?

rluders:
Well, we basically don't use any flowbite core javascript in the react-flowbite... So...

zoltanszogyenyi:
Hey peeps,

We use that to enable vertical scroll when the body content of the modal exceeds the viewport of a smaller device - it's probably not the best solution and we'll revisit this when overhauling the modal component with more options and animations.

Here's an example without the `h-modal` tag on mobile devices:

<img width="1015" alt="Screenshot 2023-01-11 at 14 27 22" src="https://user-images.githubusercontent.com/8052108/211806271-3f92bda8-8f8b-4227-8310-94401f8cff7b.png">

As you can see, we add this to make sure that the modal takes up the height of the viewport minus `2rem` to add some space between the modal and the edge of the screen.

I believe we will find a better solution when we'll update the modal component.

rluders:
@zoltanszogyenyi let me know when if got fixed at the flowbite core, so we can apply the fix here as well. :)

zoltanszogyenyi:
Hey @rluders @tulup-conner,

I've checked the behaviour of the modal on Flowbite React and we might have to generate a `<div>` that will be prepended to the last tag of the `</body>` element and apply these basic classes:

`bg-gray-900 bg-opacity-50 dark:bg-opacity-80 fixed inset-0 z-40`

The gap is there because the modal might have more content inside of it so we want to enable scroll on mobile devices so the users can navigate through all of the content available inside of the modal.

[...]

PR #658, opened 2023-03-06
fix(modal theme): fix modal vertical positioning

Ncookiez:
## Description

As of the changes of #601 (fixing #600), modals can no longer be positioned vertically. This is because the height of the modal is always 100% and thus `items-start`, `items-center` or `items-end` do not affect the flex items. This PR reverts the simple change made that broke this.

The bug can be observed in the docs [here](https://flowbite-react.com/modal) where positioning the item anywhere but the top does nothing.

In my opinion, the fix for overflowing modals seen in #601 and also suggested on #537 is not accurate. Setting a max height for your modal when using the component is more than sufficient to fix this issue on any given project, but eventually having a fix on this package could be ideal (some of this discussion overlaps with #520).

## Type of change

## How Has This Been Tested?

Manually adding `md:h-auto` to my own project and the docs' site.

## Checklist:

### DIFF
--- a/src/lib/theme/default.ts
+++ b/src/lib/theme/default.ts
@@ -541,7 +541,7 @@ const theme: FlowbiteTheme = {
       },
     },
     content: {
-      base: 'relative h-full w-full p-4',
+      base: 'relative h-full w-full p-4 md:h-auto',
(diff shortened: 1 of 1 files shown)


tulup-conner (approved):
Confirmed this behavior on the docs, thank you

Issue #612, opened 2023-02-20
Tabs automatically indexing on page load


benpatersonx:
**Describe the bug**
When you load a page using the tabs component, the first item automatically appears with a focus ring around it or 'indexed' even though it should simply just show as selected.  I believe it is an issue with the tab index.

**To Reproduce**

1. Use the 'Tabs' component from flowbite-react (src/lib/components/Tab/Tabs.tsx)
2. Load/reload the page
3. You'll see that the first tab item is automatically focused/indexed (see screenshots)

**Expected behavior**
All tabs should appear the same until the user interacts with them, apart from the first which should simply be selected not focused like it is.

**Screenshots**
![image](https://user-images.githubusercontent.com/70548361/220189498-04a676c1-ada2-4c68-ab51-55d752cef99c.png)

**System information:**

- Device: Desktop PC
- Resolution: 2560 x 1440
- OS: Windows
- Browser: Google Chrome
- Version: 109.0.5414.120

**Project information:**

- Tailwind: 3.2.7
- Flowbite: 1.6.3
- Flowbite React: 0.3.8
- Type: NextJS

evanmays:
+1

timomedia:
+1

timomedia:
Please help us fix this. Currently if each page uses tabs then when accessing the default url it will select the first tab and automatically jump the scroll bar there. It's exactly like the picture above that @benpatersonx posted.

ex-oleh:
focus ring is not the worst thing about this behavior. if you have a card component that has tabs at the bottom of the screen,
the view will always scroll to it on page refresh.

timomedia:
@ex-oleh 
Yes.
It will always scroll to the section tab when the page loads. This is a pretty bad thing.

PR #661, opened 2023-03-08
Edit default.ts

paghar:


tulup-conner:
This is not compliant with WCAG 2.2 Focus Visible (https://www.w3.org/WAI/WCAG22/Understanding/focus-visible.html).

I'm assuming that what you are trying to do is resolve #612 but this isn't an appropriate solution. The problem is that the first tab is given focus on the page by force on page load, which should be changed (https://github.com/themesberg/flowbite-react/blob/main/src/lib/components/Tab/Tabs.tsx#L115). If you're interested, feel free to open a PR with that change instead

tulup-conner:
Also, in the future please provide a more detailed title and actually provide a description for your PR or it is going to be closed automatically per the guide. This makes things easier for everyone involved.

PR #642, opened 2023-03-03
Remove `create-react-app` development server

tulup-conner:
Finally, it's time. See #221 

https://flowbite-react.vercel.app

PR #672, opened 2023-03-24
chore(pre-commit): simplify pre-commit logic to only run on changed files

### LINKED ISSUE #671: Chore: Simplify pre-commit logic
Right now the project runs this command on pre-commit: `yarn prettier:fix && yarn lint && yarn test run && yarn build`. This performs:
- Prettier on the whole project.
- Eslint on the whole project.
- Runs all of the tests.
- Fully re-builds the whole project.

There are ways to target just the changed files, like with the already installed `lint-staged` dependency. It would be nice to utilize this tool so that code can be committed more quickly. Especially as the project continues to grow, the pre-commit logic as-is will keep taking longer and longer.

Make it faster and easier to commit changes to the project. `lint-staged` is a great solution for this, and is already installed but not properly configured. I think switching the `pre-commit` husky hook to use `lint-staged` would be a great and straightforward approach.

 I'm not personally attached to exactly what commands are run as part of the `pre-commit` hook, but think it makes sense to try and avoid building the whole project for each commit, and try to stick to commands that can target only the relevant files. Ultimately, what command are run are up to you all though :).

The pre-commit hook could be left as-is, removed, or simplified to only run the linter and prettier. The pull requests are already set up to run full builds, and all of the tests I believe, so every commit does not necessarily need to do all of that checking as well.

I think `lint-staged` is easy enough to set up though, so it's worth spending a bit of time to do it. Giving quick direct feedback on issues while trying to commit can be quite helpful!

**Additional context**
I'm happy to work on this and at least create an initial PR! Feel free to give any feedback on exactly what commands you all would like as part of the lint-staged config, or if you'd prefer a different approach than lint-staged.

JackRobards:
Configure the project to use lint-staged for the pre-commit logic, instead of having lint-staged installed but not using it.

This should speed up development, and make it easier to commit changes to the project. There is less time waiting for the whole project to build or for all of the tests to run, when they are not related to your changes.

The new pre-commit workflow is:
- eslint and prettier will run one after another, instead of in parallel. This is to prevent race conditions with potentially modifying the same file: https://github.com/okonet/lint-staged#task-concurrency
- Recommended vitest options for lint-staged: https://vitest.dev/guide/cli.html#vitest-related
- tsc-files was added as a devDependency, in order to only run tsc against the changed files, instead of all files: https://github.com/gustavopch/tsc-files

Also the `lint-staged.js` file was renamed since it was not detectable by lint-staged so it did not work. The new name is the one suggested by lint-staged for use with the `module.exports` syntax: https://github.com/okonet/lint-staged#configuration

Fixes #671

## Type of change

- [X] This change contains a build config change.

## How Has This Been Tested?

- [X] Tested commiting typescript files and running `yarn lint-staged` against them. Confirmed the linter, prettier, vitest, and tsc only targeted relevant files.
- [X] Tested commiting CSS files and running `yarn lint-staged` against them. Confirmed the linter, prettier, vitest, and tsc only targeted relevant files.

## Checklist:

- [X] I have commented my code, particularly in hard-to-understand areas

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-yarn prettier:fix && yarn lint && yarn test run && yarn build
+yarn lint-staged
--- a/lint-staged.config.cjs
+++ b/lint-staged.config.cjs
@@ -1,0 +1,9 @@
+module.exports = {
+  '*.{js,jsx,ts,tsx}': [
+    'eslint --cache --fix',
+    'prettier --write',
+    'yarn test related --run',
+    () => 'tsc-files --noEmit',
+  ],
+  '*.{json,css}': ['prettier --write'],
+};
--- a/lint-staged.js
+++ b/lint-staged.js
@@ -1,8 +1,0 @@
-module.exports = {
-  '*.{js,jsx,ts,tsx}': [
-    'eslint --cache --fix',
-    'react-scripts test --bail --watchAll=false --findRelatedTests --passWithNoTests',
-    () => 'tsc -p tsconfig.json --noEmit',
-  ],
-  '*.{js,jsx,ts,tsx,json,css,js}': ['prettier --write'],
-};
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     "start-server-and-test": "^2.0.0",
     "storybook-dark-mode": "^2.1.1",
     "tailwindcss": "^3.2.7",
+    "tsc-files": "^1.1.3",
     "typescript": "^4.9.5",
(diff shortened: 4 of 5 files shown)


tulup-conner:
Thanks for the submission, but I dropped the pre commit checks in #642 entirely. I find them obnoxious because I frequently rebase to cleanup history. In other words, I agree with you, and once I merge that we won't have any pre commit hooks, and a merge is on the way shortly. I think between the documentation and the fact that your PRs get rejected by the CI/CD there's no reason to add the additional friction. If that proves to cause quality to deteriorate I love your solution instead.

JackRobards:
I didn't notice that PR - it looks great! And thanks for the detailed response 😄

PR #697, opened 2023-04-12
feat(components/breadcrumb): allows custom component to BreadcrumbItem

TungSC:
## Description
- In case developer wants to use react router for breadcrumb, we should allow dev custom component

## Type of change

## Checklist:


src/lib/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -18,12 +18,13 @@ export interface BreadcrumbItemProps extends PropsWithChildren<Omit<ComponentPro
   href?: string;
   icon?: FC<ComponentProps<'svg'>>;
   theme?: DeepPartial<FlowbiteBreadcrumbItemTheme>;
+  as?: ElementType;
```
rluders, line 21: I'm stupid. Why use `ElementType` here?

rluders: Reading the docs. I got it now.

src/lib/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -18,12 +18,13 @@ export interface BreadcrumbItemProps extends PropsWithChildren<Omit<ComponentPro
   href?: string;
```
rluders, line 18: Well, I think that I should also be able to use some components from `react-router-dom` here, right? So... For example, if I'm using the `Link` component, it needs to get the `to=` not the `href=` but, if I'm using the `a` tag it needs to support the `href=`.

I tested locally, and unless I'm doing something wrong, it doesn't seems to work.

rluders (changes requested):

src/lib/components/Breadcrumb/BreadcrumbItem.tsx
TungSC: In case I use `Link` from `next/link` it is suitable, I dont want to pass {...props} into <Component> cuz <li> used, so I'll add option `to`

rluders: I see, but I'm not sure if this is the right fix, maybe we need to introduce some breaking change here. And I'm so sorry to bring this discussion or even request to your PR... I know that it should be a simple contribution, but I think that we can explore it a little bit. WDYT?

I've been considering to allowing users to use something like:

```ts
<Breadcrumb.Item {...itemProps}><Link to="/the/page">Something here</Link></Breadcrum.Item>
```

Looks more verbose? Yes! Looks simpler? Maybe. Flexible? Yes!

But there is a catch to doing it, 'cause we may need to change the `Breadcrumb.Item` theme to add the style to the child object, or to inject the theme to the `children`. Not sure I was able to make it clear... but basically it means that:

```ts
  const Component = // Check if the child is a component and use it as a component, otherwise use some standard tag (span?)
  return (
      <li className={classNames(theme.base, className)} {...props}>
        <HiOutlineChevronRight aria-hidden className={theme.chevron} data-testid="flowbite-breadcrumb-separator" />
        <Component
          ref={ref as never}
          className={theme.href[isLink ? 'on' : 'off']}
          data-testid="flowbite-breadcrumb-item"
        >
           // how can we inject the icon here?
          {Icon && <Icon aria-hidden className={theme.icon} />}
          // here the children must be the children for the Component, not from Breadcrum.Item
          {children}          
        </Component>
      </li>
    );
```

rluders: WDYT? Does it make sense? Am I going to crazy about it?

rluders: @tulup-conner help me here. :)

tulup-conner: I agree with everything you said and also recognize the issue where we do need to maintain control over elements to apply the theme by default.

In this case I think it might just be as simple as splitting the child link component to `Breadcrumb.ItemLink`. Control over theme is retained, and the user can pass props to both child and parent independently.

TungSC: While walking around, I found this component that is used in the same way as [SidebarItem](https://github.com/themesberg/flowbite-react/blob/main/src/lib/components/Sidebar/SidebarItem.tsx). However, if you use it like this `<Breadcrumb.Item {...itemProps}><Link to="/the/page">Something here</Link></Breadcrum.Item>` may cause confusion for users to use library

rluders:
```ts
<Breadcrumb 
    renderItem={(theme, item) => <span className={theme}>item.children</span>}
    renderLink={(theme, item) => <Link {...item.props} className={theme}>{item.children}</Link>}
    {...props}
>
    <Breadcrumb.Item href="/">Home</Breadcrumb.Item>
    <Breadcrumb.Item href="/blog">Blog</Breadcrumb.Item>
    <Breadcrumb.Item>My post</Breadcrumb.Item>
</Breadcrumb>
```

🤔

PR #700, opened 2023-04-14
fix(component/models): prevent scrolling of body when modal is shown

### LINKED ISSUE #604: Prevent scrolling on body when Modal is open
Add option to prevent scrolling on `document.body` when Modal is open (or just enable it without having an actual option to do so). If the body can scroll, this leads to double scrollbars when the modal content is long.

This can be accomplished by adding the following to the Modal component:

```jsx
useEffect(() => {
  if (document) {
    document.body.style.overflow = show ? 'hidden' : 'auto'
  }
}, [show])
```

multiwebinc:
Fixes #604

### DIFF
--- a/src/lib/components/Modal/Modal.tsx
+++ b/src/lib/components/Modal/Modal.tsx
@@ -95,6 +95,13 @@ const ModalComponent: FC<ModalProps> = ({
       onClose();
     }
   });
+  
+  // Prevent scrolling of the root element when the modal is shown
+  useEffect(() => {
+    if (root) {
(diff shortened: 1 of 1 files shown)


rgwebcode:
Maybe I'm doing this not the right way, but this fails if there are multiple modals on a page. It's being triggered for each modal, and the overflow setting will be applied for the last modal in the queue (which most of the time is not the one being opened, so "auto" instead of "hidden" is being set last).

multiwebinc:
@rgwebcode You're right. It doesn't take into account multiple modals. I recommend opening an issue if you haven't done so already.

PR #703, opened 2023-04-18
feat(sidebar): Allow to customize the Sidebar component with 'as'

mvaled:
## Description

In some cases, the sidebar is the main navigation of an app and setting using

```
    <Sidebar as="nav">
       ...
    </Sidebar>
```

would be more appropriate.

## Type of change

### DIFF
--- a/src/lib/components/Sidebar/Sidebar.tsx
+++ b/src/lib/components/Sidebar/Sidebar.tsx
@@ -1,5 +1,5 @@
 import classNames from 'classnames';
-import type { ComponentProps, FC, PropsWithChildren } from 'react';
+import type { ComponentProps, ElementType, FC, PropsWithChildren } from 'react';
 import type { DeepPartial } from '..';
 import { mergeDeep } from '../../helpers/mergeDeep';
 import type { FlowbiteBoolean } from '../Flowbite/FlowbiteTheme';
@@ -31,6 +31,7 @@ export interface FlowbiteSidebarTheme {
 }
 
 export interface SidebarProps extends PropsWithChildren, ComponentProps<'div'> {
+  as?: ElementType;
   collapseBehavior?: 'collapse' | 'hide';
   collapsed?: boolean;
   theme?: DeepPartial<FlowbiteSidebarTheme>;
@@ -38,6 +39,7 @@ export interface SidebarProps extends PropsWithChildren, ComponentProps<'div'> {
 
 const SidebarComponent: FC<SidebarProps> = ({
   children,
+  as: Component = 'aside',
   collapseBehavior = 'collapse',
   collapsed: isCollapsed = false,
   theme: customTheme = {},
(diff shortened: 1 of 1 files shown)


mvaled:
@rluders 

I'd rather have the sidebar completely as the `nav`, because the main purpose of (some) sidebars is the navigation.  Non-links elements are almost always required; and they are not forbidden. 

The [first example of `<nav>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav) in MDN contains a tree of elements `ol > li > a`.   (It even state that the links could be provide in prose 🤓 .)

On the other hand, an `aside` is a semantically anything (not just links and doesn't require items, groups, etc) that is indirectly related to the main content.  

To be honest, I was somewhat surprised to see an `aside` as the default choice for a sidebar, for which the main purpose seems to be navigation (`Sidebar.Item` being mostly a link).  But I didn't want the PR to be a breaking change.

The app I'm doing has a vertical navigation as a sidebar (like Sentry's navbar):

Screenshot from 2023-04-18 09-19-10

Maybe I could use `Navbar` instead, but that requires more work because of the default horizontal styling.

rluders:
Well, @mvaled... it seems that you are correct. I wasn't aware of it, thanks for the explanation. So, in order to be more semantic-correct we should even avoid using the `aside` for the sidebar at all. Correct? So, in this case, my suggestion is to replace line `42` to use `nav` as default and allow people to use some other tags like `div`, `section`, or `aside` as they prefer.

WDYT? Does it make sense?

mvaled:
I'll include the 'nav' as the default.  Should I also include a (soft) breaking-change in the changelog.  Or do you prefer to collect changelogs at the end of the release?

rluders:
@mvaled could you check the unit tests, please? :)

tulup-conner:
I think you can argue either way on this semantically quest easily. Sidebars often include navigation but aren't limited to it. Example: https://flowbite.com/react-admin-dashboard-pro/preview/

I do agree that `nav` is more specific, though, when navigation is all the box is used for. And I can't speak for the users of the library, but I do imagine you're right that that's usually the case.

And thank you for contributing, once the unit tests are ready we can merge :D

src/lib/components/Sidebar/Sidebar.tsx
```diff
@@ -48,14 +50,14 @@ const SidebarComponent: FC<SidebarProps> = ({
 
   return (
     <SidebarContext.Provider value={{ isCollapsed }}>
-      <aside
-        aria-label="Sidebar"
+      <Component
+        aria-label="Navigation Sidebar"
```
tulup-conner, line 54: If this is default `nav`, labelling it with navigation is redundant to screen readers, which will announce the section as a navigation with label navigation sidebar. See redundant descriptions: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/navigation_role#best_practices

tulup-conner (changes requested):

mvaled:
@rluders The tests pass now.  It was the change in the `aria-label`, which was reverted due to the suggestion of @tulup-conner.

rluders:
Thank you! Let's get it merged.

rluders (approved):

PR #730, opened 2023-05-03
feat(card): support custom renderImage functions for Card

### LINKED ISSUE #706: Use NextJS image optimization
When I use the `Card` with an image, the rendered html contains a plain old `img` tag. Since I am using `tailwind-react` in a `NextJS` project, it should however use the `Image` component from `next` to benefit from image optimization.

I suggest to be able to provide a `tag` property to `Card` which defaults to `img`. Then one can overwrite it in `NextJS`-apps with `Image`.

Would also possible to make a `flowbyte-nextjs` package or something, but I think above solution is leaner.

levino:
Since in some cases, for example next.js apps, one needs to use a specific and optimized `Image` component, the `Card` component now accepts a `renderImage` property to render your own component. If `renderImage` is set, `imgSrc` and `imgAlt` are ignored. TypeScript will error if the user attempts to set them at the same time.

fix #706 

To do:

- [x] Add a test with falsy `imgSrc` -> should not render the an image tag
- [x] Update documentation.

### DIFF
--- a/app/docs/components/card/card.mdx
+++ b/app/docs/components/card/card.mdx
@@ -67,6 +67,35 @@ Add an image to the card by using the `imgSrc` prop and set the `imgAlt` prop to
   </Card>
 </CodePreview>
 
+## Card with custom image render function
+
+Specify your own render function for the image component for the card by using the `renderImage` prop. This is especially useful when
+using the component with NextJS or Gatsby.
+
+<CodePreview
+  importFlowbiteReact="Card"
+  importExternal="import Image from 'next/image';"
+  title="Card with decorative image"
+  className="max-w-sm"
+  code={`<Card renderImage={() => {<Image width={500} height={500} src="/images/blog/image-1.jpg" />}}>
+  <h5 className="text-2xl font-bold tracking-tight text-gray-900 dark:text-white">
+  Noteworthy technology acquisitions 2021
+  </h5>
+  <p className="font-normal text-gray-700 dark:text-gray-400">
+  Here are the biggest enterprise technology acquisitions of 2021 so far, in reverse chronological order.
+  </p>
+</Card>`}
+>
+  <Card renderImage={() => <Image width={500} height={500} src="/images/blog/image-1.jpg" />}>
+    <h5 className="text-2xl font-bold tracking-tight text-gray-900 dark:text-white">
+      Noteworthy technology acquisitions 2021
+    </h5>
+    <p className="font-normal text-gray-700 dark:text-gray-400">
+      Here are the biggest enterprise technology acquisitions of 2021 so far, in reverse chronological order.
+    </p>
+  </Card>
+</CodePreview>
+
 ## Horizontal card
 
 Use the `horizontal` prop to show the card in a horizontal layout.
--- a/src/components/Card/Card.tsx
+++ b/src/components/Card/Card.tsx
@@ -3,6 +3,7 @@ import { twMerge } from 'tailwind-merge';
 import type { DeepPartial, FlowbiteBoolean } from '../../';
 import { useTheme } from '../../';
 import { mergeDeep } from '../../helpers/merge-deep';
+import { omit } from '../../helpers/omit';
 
 export interface FlowbiteCardTheme {
   root: FlowbiteCardRootTheme;
@@ -21,26 +22,32 @@ export interface FlowbiteCardImageTheme {
   horizontal: FlowbiteBoolean;
 }
 
-export interface CardProps extends PropsWithChildren<ComponentProps<'div'>> {
+interface CommonCardProps extends PropsWithChildren<ComponentProps<'div'>> {
   horizontal?: boolean;
   href?: string;
-  imgAlt?: string;
-  imgSrc?: string;
+  /** Overwrites the theme. Will be merged with the context theme.
+   * @default {}
+   */
   theme?: DeepPartial<FlowbiteCardTheme>;
 }
 
-export const Card: FC<CardProps> = ({
-  children,
-  className,
-  horizontal,
-  href,
-  imgAlt,
-  imgSrc,
-  theme: customTheme = {},
-  ...props
-}) => {
+export type CardProps =
+  | (
+      | { imgAlt?: string; imgSrc?: string; renderImage?: never }
+      | {
+          /** Allows to provide a custom render function for the image component. Useful in Next.JS and Gatsby. **Setting this will disable `imgSrc` and `imgAlt`**.
+           */
+          renderImage?: (theme: DeepPartial<FlowbiteCardTheme>, horizontal: boolean) => JSX.Element;
+          imgAlt?: never;
+          imgSrc?: never;
+        }
+    ) &
+      CommonCardProps;
+
+export const Card: FC<CardProps> = (props) => {
+  const { children, className, horizontal, href, theme: customTheme = {} } = props;
   const Component = typeof href === 'undefined' ? 'div' : 'a';
-  const theirProps = props as object;
+  const theirProps = removeCustomProps(props);
 
   const theme = mergeDeep(useTheme().theme.card, customTheme);
 
@@ -56,16 +63,38 @@ export const Card: FC<CardProps> = ({
       )}
       {...theirProps}
     >
-      {imgSrc && (
-        <img
-          alt={imgAlt ?? ''}
-          src={imgSrc}
-          className={twMerge(theme.img.base, theme.img.horizontal[horizontal ? 'on' : 'off'])}
-        />
-      )}
+      {/* eslint-disable-next-line jsx-a11y/alt-text -- jsx-ally/alt-text gives a false positive here. Since we use our own Image component, we cannot provide an "alt" prop.*/}
+      <Image {...props} />
       <div className={theme.root.children}>{children}</div>
     </Component>
   );
 };
 
-Card.displayName = 'Card';
+const Image: FC<CardProps> = ({ theme: customTheme = {}, ...props }) => {
+  const theme = mergeDeep(useTheme().theme.card, customTheme);
+  if (props.renderImage) {
+    return props.renderImage(theme, props.horizontal ?? false);
+  }
+  if (props.imgSrc) {
+    return (
+      <img
+        data-testid="flowbite-card-image"
+        alt={props.imgAlt ?? ''}
+        src={props.imgSrc}
+        className={twMerge(theme.img.base, theme.img.horizontal[props.horizontal ? 'on' : 'off'])}
+      />
+    );
+  }
+  return null;
+};
+
+const removeCustomProps = omit([
+  'renderImage',
+  'imgSrc',
+  'imgAlt',
+  'children',
+  'className',
+  'horizontal',
+  'href',
+  'theme',
+]);
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,6 +2,11 @@
 # yarn lockfile v1
 
 
+"@aashutoshrathi/word-wrap@^1.2.3":
+  version "1.2.6"
+  resolved "https://registry.yarnpkg.com/@aashutoshrathi/word-wrap/-/word-wrap-1.2.6.tgz#bd9154aec9983f77b3a034ecaa015c2e4201f6cf"
+  integrity sha512-1Yjs2SvM8TflER/OD3cOjhWWOZb58A2t7wpE2S9XfBYTiIl+XFhQG2bjy4Pu1I+EAlCNUzRDYDdFwFYUKvXcIA==
+
 "@adobe/css-tools@^4.0.1":
   version "4.2.0"
   resolved "https://registry.yarnpkg.com/@adobe/css-tools/-/css-tools-4.2.0.tgz#e1a84fca468f4b337816fcb7f0964beb620ba855"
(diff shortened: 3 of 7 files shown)


src/lib/components/Card/Card.stories.tsx
```diff
@@ -45,3 +46,17 @@ WithDecorativeImage.storyName = 'With decorative image';
...
   imgSrc: 'https://flowbite.com/docs/images/blog/image-1.jpg',
 };
+
+export const WithNextImage = Template.bind({});
+WithNextImage.storyName = 'With Next.js Image component';
+WithNextImage.args = {
+  renderImage: () => (
+    <Image
+      alt="Meaningful alt text for an image that is not purely decorative"
+      loader={({ src }) => src}
```
levino, line 56: The loader prop here is important. If we do not provide this, `Image` returns some weird "ExoticComponent" and the Storybook crashes.

src/lib/components/Card/Card.tsx
```diff
@@ -22,29 +23,40 @@ export interface FlowbiteCardImageTheme {
...
+type CommonCardProps = PropsWithChildren<ComponentProps<'div'>> & {
   horizontal?: boolean;
   href?: string;
-  imgAlt?: string;
-  imgSrc?: string;
   theme?: DeepPartial<FlowbiteCardTheme>;
-}
+};
+
+export type CardProps =
```
levino, line 32: I am not sure if it is possible to create XOR types with "interfaces". If you feel strongly about the usage of "type" instead of "interface" I can try to improve this. However I then also recommend adding a linter rule which disallows the usage of "type".

src/lib/components/Card/Card.tsx
```diff
@@ -22,29 +23,40 @@ export interface FlowbiteCardImageTheme {
...
-  href,
-  imgAlt,
-  imgSrc,
-  theme: customTheme = {},
-  ...props
-}) => {
+export const Card: FC<CardProps> = ({ children, className, horizontal, href, theme: customTheme = {}, ...props }) => {
   const Component = typeof href === 'undefined' ? 'div' : 'a';
-  const theirProps = props as object;
+  const theirProps = omit(props, ['renderImage', 'imgSrc', 'imgAlt']) as object;
```
levino, line 45: I find the gather above somewhat dangerous to begin with. I clean the props up here a bit by at least removing, what I do not want to have in the `Component` for sure.

src/lib/components/Card/Card.tsx
```diff
@@ -22,29 +23,40 @@ export interface FlowbiteCardImageTheme {
...
+  const Image = () =>
+    'renderImage' in props && props.renderImage ? (
+      props.renderImage()
+    ) : 'imgSrc' in props ? (
+      <img
+        alt={props.imgAlt ?? ''}
+        src={props.imgSrc}
+        className={classNames(theme.img.base, theme.img.horizontal[horizontal ? 'on' : 'off'])}
+      />
+    ) : null;
```
levino, line 58: I am not such a big fan of such ternary magic. One could pull this out into a couple of new components but then also that might get a little out of hand. Please advise.

src/lib/helpers/omit.ts
```diff
@@ -0,0 +1,10 @@
+export const omit = <T extends Record<string, unknown>, K extends keyof T>(obj: T, keys: string[]): Omit<T, K> =>
```
levino, line 1: I could not find `omit` or any useful helper library in the codebase (like `lodash`, `underscore` or `ramda`) so I just added this.

levino (commented):
This is a first draft. Please have a look and let me know what you like and what not.

src/lib/components/Card/Card.tsx
texonidas: You could achieve this via null coalescing instead? Also you've changed the behaviour by doing a check if imgSrc is present, when before it was only checking if it was truthy, which I have rolled back.

```suggestion
  const Image = () => props.renderImage?.() ??  props.imgSrc ? (
      <img
        alt={props.imgAlt ?? ''}
        src={props.imgSrc}
        className={classNames(theme.img.base, theme.img.horizontal[horizontal ? 'on' : 'off'])}
      />
    ) : null;
```

levino: Did you check that this works with typescript? About the `imgSrc`: Are you saying that now it would render an image, if someone set `imgSrc` to `false` and that is a problem? Maybe one should add a test for that... I will give it some love and consider your remarks.

Overall I have to say this is very, very hard to read and make sense of and the coalescing is "cool" but adding mental load imo.

levino:
I will continue to work on this later. I will also update the docs.

src/lib/helpers/omit.ts
```diff
@@ -0,0 +1,14 @@
+export const omit =
+  <T extends object, K extends string>(keys: readonly K[]) =>
+  (obj: T): Omit<T, K> => {
```
levino, line 3: I decided to curry this and change the order of the arguments since that is easier to use to create partially applied functions.

src/lib/components/Card/Card.stories.tsx
```diff
@@ -1,17 +1,12 @@
...
   title: 'Components/Card',
   component: Card,
-  decorators: [
-    (Story): JSX.Element => (
-      <div className="h-1/2 w-1/2">
-        <Story />
-      </div>
-    ),
-  ],
+  decorators: [(Story): JSX.Element => <div className="h-1/2 w-1/2">{Story()}</div>],
```
levino, line 9: This change was important. This fixes the broken code previews in the storybook docs:
Screenshot 2023-05-04 at 14 32 37

src/lib/components/Card/Card.tsx
```diff
@@ -57,14 +64,37 @@ export const Card: FC<CardProps> = ({
...
+
+const Image: FC<CardProps> = ({ theme: customTheme = {}, ...props }) => {
+  const theme = mergeDeep(useTheme().theme.card, customTheme);
+  if (props.renderImage) {
+    return props.renderImage();
+  }
+  if (props.imgSrc) {
+    return (
+      <img
+        data-testid="flowbite-card-image"
```
levino, line 81: Usage of testId is a little uncool, but I could find a better way to check that "this component is not rendered" from the tests.

levino:
Can someone please review my PR? Would be so sad if all the work was in vain...

src/lib/components/Card/Card.tsx
rluders: This is outdate, so, I'll mark it as resolved.

src/lib/components/Card/Card.tsx
rluders: This is outdate, so, I'll mark it as resolved.

src/lib/components/Card/Card.stories.tsx
rluders: This is only to satisfy the storybook, right?

src/lib/components/Card/Card.tsx
rluders: I would always prefer to use `Interface`. Can you change it?

src/lib/components/Card/Card.stories.tsx
```diff
@@ -45,3 +40,17 @@ WithDecorativeImage.storyName = 'With decorative image';
 WithDecorativeImage.args = {
   imgSrc: 'https://flowbite.com/docs/images/blog/image-1.jpg',
 };
+
+export const WithNextImage = Template.bind({});
+WithNextImage.storyName = 'With Next.js Image component';
+WithNextImage.args = {
+  renderImage: () => (
```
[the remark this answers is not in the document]
rluders: If I'm not wrong at the issue we had agreed to use `imageAs`... don't we?

src/lib/components/Card/Card.tsx
```diff
@@ -57,14 +59,36 @@ export const Card: FC<CardProps> = ({
...
+      <Image {...props} />
       <div className={theme.root.children}>{children}</div>
     </Component>
   );
 };
+
+const Image: FC<CardProps> = ({ theme: customTheme = {}, ...props }) => {
+  const theme = mergeDeep(useTheme().theme.card, customTheme);
+  if (props.renderImage) {
+    return props.renderImage();
```
rluders, line 76: We need to find a way to properly pass the theme classes to the rendered image. Otherwise, this tag will not be properly styled to flowbite.

src/lib/components/Card/Card.tsx
```diff
@@ -57,14 +59,36 @@ export const Card: FC<CardProps> = ({
...
+  }
+  if (props.imgSrc) {
+    return (
+      <img
+        alt={props.imgAlt ?? ''}
+        src={props.imgSrc}
+        className={classNames(theme.img.base, theme.img.horizontal[props.horizontal ? 'on' : 'off'])}
```
rluders, line 84: This is the theme that I'm talking about. :)

yarn.lock
```diff
@@ -12794,6 +12868,28 @@ nested-error-stacks@^2.0.0, nested-error-stacks@^2.1.0:
   resolved "https://registry.yarnpkg.com/nested-error-stacks/-/nested-error-stacks-2.1.1.tgz#26c8a3cee6cc05fbcf1e333cd2fc3e003326c0b5"
   integrity sha512-9iN1ka/9zmX1ZvLV9ewJYEk9h7RyRRtqdK0woXcqohu8EWIerfPUjYJPg0ULy0UqP7cslmdGc8xKDJcojlKiaw==
 
+next@^13.3.4:
```
rluders, line 12871: I'm thinking if we should have NextJS as a dependency here. WDYT, @tulup-conner?

rluders (changes requested):
Looks great. I requested a few changes, and I would also love to listen to @tulup-conner's opinion on this one.

src/lib/components/Card/Card.stories.tsx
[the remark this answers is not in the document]
levino: It is not a component, but a render function like for example with [formik](https://formik.org/docs/api/formik#render-props-formikpropsvalues--reactnode). Something like `imageAs="div"` will not work. Also I will not inject `imgSrc` or `imgAlt` into this function. The function will receive nothing, you have to bind whatever you want in the context. I think it is just too complicated to expect the user to pass through some of the variables and others not (like `width` and `height` with `Image` from `nextjs`).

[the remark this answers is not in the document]
rluders: @levino I see your point and I totally agree with you. Can you check my comment at #734?

src/lib/components/Card/Card.tsx
rluders: The flowbite-react theme system allows you to centralize the style of the components in one single place. So, it means that all your Cards will first use the style from the default theme unless you decide to overwrite it. It is necessary from the library perspective that all things inside the components should be accessible by the theme properties to be stylized. It is a way that we found to keep the design system consistency, but it also allows exceptions (when really necessary).

src/lib/components/Card/Card.tsx
rluders: But you should have access to the theme, and you can decide if you will use it or not.

src/lib/components/Card/Card.tsx
levino: It is an image. What do you want to style with an image?

src/lib/components/Card/Card.tsx
levino: Yes, just do:

```
renderImage = () => {
  const theme = useTheme()
  ....
}

File: src/lib/components/Card/Card.tsx (hunk 10 above)
Comment (rluders) [reply src/lib/components/Card/Card.tsx:76]: [Well, first that we already ](https://github.com/themesberg/flowbite-react/blob/89d58dc03727d150e73ec753748eb98640bc84a3/src/lib/theme/default.ts#L332)

☝️ Cards can be horizontal or vertical layouts, and it impacts how the image will be styled.

Comment (levino) [reply src/lib/components/Card/Card.tsx:76]: Maybe the "custom image component" is a `div` which gets a background image. Maybe someone wants to render a `form`. Who knows what kind of hacky optimization magic the user might be doing. We cannot foresee what they come up with. Better tell them: You are on your own, no magic provided on our end.

File: src/lib/components/Card/Card.tsx (hunk 11 above)
Comment (rluders) [reply src/lib/components/Card/Card.tsx:84]: TBH, I don't see the problem to inject, and I don't see the benefits to call the hook again.

I got the point that "one unified way to access the theme", and it seems a pretty reasonable argument, but since we are using the `renderImage` to render something exclusively inside the `Card` component it kind of loses its point. I already have the theme object available inside the component where I'm trying to render an "expected" component. So, properties that can be used in this inner component should be injected by the main component.

Also, it will make the code a little bit clean... Also consider that, if you are rendering a list of cards, and then you have 10 cards in this list. You are going to call the hooks 20 times. If you inject, you will have only half (not ideal yet, but, still... half).

File: src/lib/components/Card/Card.tsx (hunk 10 above)
Comment (rluders) [reply src/lib/components/Card/Card.tsx:76]: I think this discussion is overlapping with the one that we are having in the other component. Let's keep it there... I'm happy with `renderImage` property.

File: src/lib/components/Card/Card.tsx (hunk 11 above)
Comment (rluders) [reply src/lib/components/Card/Card.tsx:84]: The theme is always present, it is part of the core:

https://github.com/themesberg/flowbite-react/blob/89d58dc03727d150e73ec753748eb98640bc84a3/src/lib/components/Card/Card.tsx#L46

Here is the interface for the card image theme:

https://github.com/themesberg/flowbite-react/blob/89d58dc03727d150e73ec753748eb98640bc84a3/src/lib/components/Card/Card.tsx#L20

Comment (rluders) [reply src/lib/components/Card/Card.tsx:84]: To the `renderImage` param you need to pass: `renderImage(theme.image)`

Comment (rluders) [reply src/lib/components/Card/Card.tsx:84]: Ops... Sorry:

```ts
interface RenderImage {
    (theme: FlowbiteCardImageTheme) => JSX.Element
}
```

rluders: You may also need to pass one extra property, let me check...

src/lib/components/Card/Card.stories.tsx
levino: No. This is a hack to make `Image` from `next` actually return a `ReactNode`. If you do not provide this, `Image` will return some funky object that makes the code crash. I guess it has something to do with an additional build step with `nextjs` where `nextjs` picks up these objects and finalizes them by injecting the proper loader. If we provide it here, everything works and the image gets rendered (without any optimizations of course). I will add a comment to explain this funkyness.

levino:
```suggestion
      {/* Necessary to make `Image` render a React Node */}
      loader={({ src }) => src}
```

levino:
Just FYI: The docs have become somewhat ugly (at least storybook / docs):
Screenshot 2023-05-10 at 11 01 42

src/lib/components/Card/Card.stories.tsx
```diff
@@ -45,3 +40,18 @@
...
+    <Image
+      alt="Meaningful alt text for an image that is not purely decorative"
+      {/* Necessary to make `Image` render a React Node */}
```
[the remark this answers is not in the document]
levino: Fixed.

levino:
Once this has been accepted, I will rebase it all into one clean commit.

src/lib/components/Card/Card.tsx
```diff
@@ -38,7 +38,7 @@ export type CardProps =
       | {
           /** Allows to provide a custom render function for the image component. Useful in Next.JS and Gatsby. **Setting this will disable `imgSrc` and `imgAlt`**.
            */
-          renderImage?: () => JSX.Element;
+          renderImage?: (config: { theme: DeepPartial<FlowbiteCardTheme>; horizontal: boolean }) => JSX.Element;
```
rluders, line 41: I don't think that we need this `config` object. Just passing the two parameters is good enough, otherwise, if you want to use the config object, you need to create an interface for it.

rluders (changes requested):

src/lib/components/Card/Card.tsx
levino: I am not sure that I understand your point correctly. Are you saying: "In order to get proper types in the storybook docs you need to create a custom interface here."? That does not help, I fear since the types for the theme in the card props have the same poor quality:
Screenshot 2023-05-15 at 07 22 50
I think one would have to somehow define a "Theme" type alias or something for the docs feature of storybook but since this is already "broken", I do not see a point in fixing it here.
If you want me to change the interface for other reasons, I can of course do it. What do you like? This:
```suggestion
          renderImage?: (theme: DeepPartial<FlowbiteCardTheme>, horizontal: boolean) => JSX.Element;
```
?
As I said, I personally find it a bit difficult because one has to remember which position is for which value in the arguments, as opposed to one argument that has named keys. Also adding more values later might become quite tedious for users when we have 4 or 5 arguments here. But I also do not have strong opinions on it (as I will probably never use these arguments in my `renderImage` functions).

levino:
@rluders Could you please resolve all conversations that have been resolved from your point of view? Will @tulup-conner come bye any time soon to give their final review?

src/lib/components/Card/Card.tsx
levino: Please give me feed back @rluders I do not want to drag this on for too long. From experience I can say: The longer a PR stays open, the less likely it will get merged eventually.

rluders: @levino I got it, but this is not my main job. I have a lot on my plate. But I'll try to test it this week.

rluders:
@levino I think that this one is good, if you could just fix the test - it is good to be merged then.

levino:
I still need to update the "real" [documentation](https://flowbite-react-git-fork-levino-main-themesberg.vercel.app/docs/components/card) with an example. Putting this to draft until I have time to fix that, probably next week. Enjoy the week end and thank you for coming back to this @rluders

rluders:
@levino I think that this is a problem that we are having right now with the CodePreview, we are trying to fix it. So, I would say... maybe doesn't matter right now?

cglacet:
Any news on this?

app/docs/components/card/card.mdx
```diff
@@ -67,6 +67,27 @@ Add an image to the card by using the `imgSrc` prop and set the `imgAlt` prop to
...
+Specify your own render function for the image component for the card by using the `renderImage` prop. This is especially useful when
+using the component with NextJS or Gatsby.
+
+<CodePreview
+  importFlowbiteReact="Card"
+  importExternal="import Image from 'next/image'"
+  title="Card with decorative image"
+  className="max-w-sm"
+>
+  <Card renderImage={() => <Image width={500} height={500} src="/images/blog/image-1.jpg" />}>
```
levino, line 81: Unfortunately this does not work as expected but there is little I can do about this. The `CodePreview` does not properly support external dependencies.

src/components/Card/Card.tsx
```diff
@@ -1,8 +1,9 @@
 import type { ComponentProps, FC, PropsWithChildren } from 'react';
 import { twMerge } from 'tailwind-merge';
-import type { DeepPartial, FlowbiteBoolean } from '../../';
-import { useTheme } from '../../';
```
levino, line 4: Is it okay to remove these relativ imports for the `~/src` below? If not, I would be curious why.

levino:
Good to go imo.

levino (commented):
Good to go imo.

app/docs/components/card/card.mdx
rluders: @tulup-conner @zoltanszogyenyi any idea?

src/components/Card/Card.tsx
rluders: Does the `yarn build:lib` still working? If it does, no issue here...

tulup-conner: Hey @levino anything inside `src/` should just be relative paths `../../` as opposed to the `~`. Unfortunately when we use the absolute path it seems to cause issues for `flowbite-react` once it's actually shipped out on `npm`. At the moment we are retaining `tsconfig.json` paths solely to make writing the documentation in `.mdx` files easier (since your editor won't automatically write `import`s for you in this file format).

app/docs/components/card/card.mdx
levino: Great stuff. Looks better now.
Screenshot 2023-07-12 at 08 15 18

src/components/Card/Card.tsx
levino: Fixed.

levino (commented):
@rluders I fixed the outstanding issues. If you think they are resolved, please resolve the conversations. Other than that, I would say, this can go out. Enjoy!

rluders:
Amazing job, @levino. Thank you very much.

rluders (approved):

PR #731, opened 2023-05-04
feat(pagination.tsx): next and previous buttons should disable when on 1 and last page respectively

### LINKED ISSUE #726: Next and Previous buttons on Pagination should disable when on 1 and last page respectively.
Next and Previous buttons on Pagination should disable when on first and last page respectively.

When currentPage === 1 disable the Previous button
When currentPage === lastPage disable the Next button

I can override the behaviour by passing renderPaginationButton and implement the disable functionality. But I think the pagination component should support this feature.

**Additional context**
None

lokeshpathrabe:
next and previous buttons should disable when on 1 and last page respectively

## Description

The Previous buttons should disable when current page is the first page.
The Next button should disable when current page is the last page.

Fixes #726 

## Type of change

## How Has This Been Tested?

Tested in the docs site and added component test for the change.

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Firmware version: 
- Hardware:
- Toolchain:
- SDK:

## Checklist:

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any misspellings


lokeshpathrabe:
@rluders Can you please review this.

src/lib/components/Pagination/PaginationButton.tsx
```diff
@@ -23,6 +25,7 @@ const PaginationButton: FC<PaginationButtonProps> = ({
   className,
   onClick,
   theme: customTheme = {},
+  disabled,
```
rluders, line 28: `disabled` isn't the same as that `active = false`?

rluders (changes requested):

src/lib/components/Pagination/PaginationButton.tsx
rluders: Or does it have another meaning here? Can it be `active` and `disabled` at the same time?

rluders: checking the code, it seems to be possible... but my question is more about the logic itself. :)

lokeshpathrabe: @rluders  Here I want to use `disabled` state only on the Prev and Next buttons and not the other pagination buttons. Prev and Next buttons will never have `active` state. 

But since all the pagination buttons including Prev and Next use PaginationButton component, I added `disabled` state there. I understand your point that `disabled` state wont make sense for PaginationButton component.

I guess I can create a separate Button to use for Next and Previous buttons so to avoid adding `disabled` state on the PaginationButton component.

What do you suggest ?

lokeshpathrabe: @rluders I have updated the code with new changes. Please take look and let me know your opinion.

src/lib/components/Pagination/PaginationButton.tsx
```diff
@@ -44,4 +49,32 @@ const PaginationButton: FC<PaginationButtonProps> = ({
...
 
 PaginationButton.displayName = 'Pagination.Button';
-export default PaginationButton;
+
+export const PaginationNavigation: FC<PaginationPrevButtonProps> = ({
+  children,
+  className,
+  onClick,
+  theme: customTheme = {},
+  disabled,
```
rluders, line 58: Maybe `disabled = false,`?

rluders (changes requested):

src/lib/components/Pagination/PaginationButton.tsx
lokeshpathrabe: @rluders Thanks for noticing, missed it. Fixed now.

rluders (approved):

Issue #737, opened 2023-05-11
nextjs  DarkThemeToggle show Warning


fourcels:
**Describe the bug**
```
Warning: Prop `fillRule` did not match. Server: "null" Client: "evenodd"
    at path
    at svg
    at IconBase
    at HiSun
    at button
    at DarkThemeToggle (webpack-internal:///(app-client)/./node_modules/.pnpm/flowbite-react@0.4.4_flowbite@1.6.5_react-dom@18.2.0_react@18.2.0_tailwindcss@3.3.2/node_modules/flowbite-react/lib/esm/components/DarkThemeToggle/DarkThemeToggle.js:19:28)
    at Flowbite (webpack-internal:///(app-client)/./node_modules/.pnpm/flowbite-react@0.4.4_flowbite@1.6.5_react-dom@18.2.0_react@18.2.0_tailwindcss@3.3.2/node_modules/flowbite-react/lib/esm/components/Flowbite/Flowbite.js:20:21)
```
**To Reproduce**
```tsx
    <Flowbite>
      <DarkThemeToggle />
    </Flowbite>
```

**Expected behavior**

**Screenshots**

**System information:**

- Device:
- Resolution:
- OS:
- Browser:
- Version:

**Project information:**

- Tailwind: 3.3.2
- Flowbite: 1.6.5
- Flowbite React: 0.4.4
- Type: nextjs

**Additional context**
system default dark mode

rluders:
This is really weird to be happening inside the DarkThemeToggle. I see how it can be related to #738 now, but at the same time, I don't believe that the proposed solution on that PR is the right answer. We should be careful by removing features from the library, and if we think that this is the only solution, maybe first, bring it to the discussion. 

What would be the pros and cons to remove the user preference stuff from the ThemeContext? If we remove it, what would be the approach to the users that want to implement it by themselves? Also, it would be a breaking feature... So, it should be documented.

SutuSebastian:
Fixed in the latest release, [v0.7.0](https://www.npmjs.com/package/flowbite-react/v/0.7.0)

PR #738, opened 2023-05-11
fix(DarkThemeToggle): nextjs waring

fourcels:
fix issue: #737


rluders:
This is not related to the reported issue, and this is basically removing a feature for no reason.

PR #739, opened 2023-05-12
Corrected syntax error on modal

oluwaseye:
Fixed a typo, syntax error on Modal component. Breaks during build


rluders:
Well, yes... now it is breaking.

PR #758, opened 2023-05-22
feat(textinput.tsx): added renderIcon and renderRightIcon to allow for Icon onClick

### LINKED ISSUE #734: TextInput icon onClick
There is no straight way to handle TextInput's icon onClick event...example is clicking the Eye icon to unmask password input

A straight forward way of doing this, probably a prop called iconClick

Nothing yet

rtraficante:
TextInput icon onClick #734

## Description

This change allows for users to render Icons using the renderIcon function property, instead of the icon property. This gives users the ability to add custom props to their Icon, such as an onClick handler. The same goes for the renderRightIcon property, but the Icon will be displayed on the right side of the text input instead.

Fixes #734

## Type of change

- [X] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

I created 4 new tests to the TextInput.spec.tsx file that check if the Icon is present in the document and make sure the onClick handler works for both the renderIcon and renderRightIcon properties

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works


src/lib/components/TextInput/TextInput.tsx
```diff
@@ -74,16 +78,25 @@ export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
...
               <div className={theme.field.icon.base}>
                 <Icon className={theme.field.icon.svg} />
               </div>
-            )}
-            {RightIcon && (
+            ) : renderIcon ? (
+              <div data-testid="rendered-icon" className={theme.field.icon.base}>
+                {renderIcon(theme.field.icon.svg)}
+              </div>
+            ) : null}
```
rluders, line 89: I would not use `null` here, instead, I would replace the `renderIcon` with `renderIcon && (...)` - same for `renderRightIcon`

Except by that, it lgtm.

rluders (changes requested):

src/lib/components/TextInput/TextInput.tsx
rtraficante: I tried something like that and was getting some weird behavior so I just went with the way I knew would work, but I will definitely try and figure it out.

rluders: Thanks. @rtraficante.

rtraficante: No problem! @rluders

src/lib/components/TextInput/TextInput.spec.tsx
```diff
@@ -15,5 +15,63 @@ describe.concurrent('Components / Text input', () => {
...
+
+    it('rendered Icon should allow for click events', () => {
+      render(
+        <TextInput
+          renderIcon={(style) => (
+            <HiEye
+              data-testid="icon"
+              className={style}
+              onClick={() => {
+                const target = document.querySelector('[data-testid="rendered-icon"]');
```
tulup-conner, line 42: Can you alter these examples to do something with `useState`, for example? Using `createElement` and `insertBefore` aren't supported React behaviors, so it doesn't really make sense to test them here.

src/lib/components/TextInput/TextInput.spec.tsx
```diff
@@ -15,5 +15,63 @@ describe.concurrent('Components / Text input', () => {
...
+                const div = document.createElement('div');
+                div.innerHTML = 'clicked';
+                target?.parentNode?.insertBefore(div, target.nextSibling);
+              }}
+            />
+          )}
+        />,
+      );
+      const icon = screen.getByTestId('icon');
+      fireEvent.click(icon);
```
tulup-conner, line 52: Please use the syntax seen, e.g. here https://github.com/themesberg/flowbite-react/blob/main/src/components/Accordion/Accordion.spec.tsx#L69

src/lib/components/TextInput/TextInput.tsx
```diff
@@ -79,11 +83,25 @@ export const TextInput = forwardRef<HTMLInputElement, TextInputProps>(
...
+            )}
+
             {RightIcon && (
               <div data-testid="right-icon" className={theme.field.rightIcon.base}>
                 <RightIcon className={theme.field.rightIcon.svg} />
               </div>
             )}
+
+            {renderRightIcon && !RightIcon && (
+              <div data-testid="rendered-right-icon" className={theme.field.rightIcon.base}>
```
tulup-conner, line 100: A testid like `flowbite-textinput-righticon` would probably be more appropriate, but that's a nitpick, I don't really care that much if you change it.

tulup-conner (changes requested):
Thanks, this is great. A few things:

First, we literally just published a complete rewrite of the development environment, so you'll need to rebase this PR. If you can't do that, let me know and I will do so for you.

I think it would be best to just replace the `rightIcon` and `icon` props with your new approach, so you should do that here. However, I do like the naming prefix `render` here, so, I think we can just keep the new props exactly as they are and delete `icon` and `rightIcon`.

That will be a breaking change, so we'll release this with 1.0.0, and I will take care of reporting the change to users.

Also, if you could be so kind, we'll need to change the examples that use `icon` or `rightIcon`, between the documentation in `app/` folder and Storybook `.stories.tsx`. There should only be a few at most.

rtraficante:
@tulup-conner When fixing up the documentation, I noticed that when the property is a function, the function doesn't actually show in the code preview. It is showing "renderIcon={renderIcon}" instead of renderIcon={(style) => ...}. I'm not sure if this is on purpose, but I think it would be a good idea to show how to pass the theme into the icon. 

I checked the code-preview component and I'm getting an error "Property replaceAll does not exist on type string. Do you need to change your target library? Try changing the 'lib' compiler option to es2021 or later."

tulup-conner:
> @tulup-conner When fixing up the documentation, I noticed that when the property is a function, the function doesn't actually show in the code preview. It is showing "renderIcon={renderIcon}" instead of renderIcon={(style) => ...}. I'm not sure if this is on purpose, but I think it would be a good idea to show how to pass the theme into the icon.
> 
> I checked the code-preview component and I'm getting an error "Property replaceAll does not exist on type string. Do you need to change your target library? Try changing the 'lib' compiler option to es2021 or later."

Yeah, this is something I'm trying to fix. We might have to do something hacky to fix it, which is fine. I'll resolve that, though, so don't worry about it. I just created an issue so we can track this publicly #771

abbasmashaddy72:
Hi I think this feature is not adding in 0.7.2 till now as I am unable to use it

tulup-conner:
Are you still able to finish this @rtraficante?

IsabelFreitas-catapult:
any updates on when this PR will be merged?

SutuSebastian:
> any updates on when this PR will be merged?

It has loads of conflicts, maybe a re-do and a re-evaluation would be a better idea here.

SutuSebastian:
Closing due to age and inactivity.

Issue #624, opened 2023-02-23
Allow customizing the button for dropdowns


multiwebinc:
The Dropdown component is very opinionated.

I suggest adding a customizable `<Dropdown.Trigger>` component that would be used to render the `<button>` element. The `label` prop should be removed from `<Dropdown>` and `arrowIcon` and `floatingArrow` moved to `<Dropdown.Trigger>`.

Also, I would assume that the `<Dropdown.Item>` components would then need to be wrapped to differentiate between the trigger and the items.

Here's example markup that would render a red button with a 3 vertical dots icon:

```jsx
<Dropdown>
  <Dropdown.Trigger className="bg-red-500" arrowIcon={false}>
    Show the dropdown
    <HiDotsVertical />
  </Dropdown.Trigger>
  <Dropdown.Items>
    <Dropdown.Item icon={HiOutlinePencil} onClick={openRatePlanEditDialogAndCloseMenu}>
      Edit
    </Dropdown.Item>
    <Dropdown.Item icon={HiOutlineTrash} className="text-red-700">
      Delete
    </Dropdown.Item>
  </Dropdown.Items>
</Dropdown>
```

Or maybe allow the dropdown to be triggered from any element, not just buttons. Someone might want to use an icon instead of a button. For that, we could add a `<Trigger.Button>` element that would render the button with the arrow icon, but any children of `<Dropdown.Trigger>` can instead be used as a trigger.

multiwebinc:
Now I see you can do `<Dropdown inline={true}>` and then use custom JSX for a button as a label. Maybe this issue isn't an issue after all.

tulup-conner:
Yep, Dropdown is too opinionated. I like your idea - if you have the time to PR this it would be greatly appreciated

Bilalzaiter:
how to make dropdown trigger on hover instead of click

jcolvard:
I'm confused, cn you not customize a Dropdown <button> with tailwind utility class names?

How come

`<Dropdown label="Dropdown button" color="red">
      <Dropdown.Item>Dashboard</Dropdown.Item>
      <Dropdown.Item>Settings</Dropdown.Item>
      <Dropdown.Item>Earnings</Dropdown.Item>
      <Dropdown.Item>Sign out</Dropdown.Item>
    </Dropdown>`

changes the color of the button to red and leaves the dropdown menu alone, and 
    
`<Dropdown label="Dropdown button" className="bg-red-500 hover:bg-red-600">
[...]

PR #778, opened 2023-05-29
feat(dropdown): add custom trigger

paghar:
#624

## Description

Please include a summary of the change and which issue is fixed. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

## Breaking changes

Please document the breaking changes if suitable.

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

- Firmware version:
- Hardware:
- Toolchain:
- SDK:

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] My changes generate no new warnings
- [ ] I have checked my code and corrected any misspellings


paghar:
@rluders I'm not sure I got the issue concept correctly. My solution is right. If it is ok, I will add a test case later.

app/docs/components/dropdown/dropdown.mdx
```diff
@@ -123,3 +123,11 @@ The dropdown component can be used to show a list of menu items when clicking on
     </div>
   </div>
 </CodePreview>
+<CodePreview title="Custom trigger dropdown">
+  <Dropdown dismissOnClick={false} CustomTrigger={<button>Custom Dropdown button</button>}>
```
rluders, line 127: I think that this approach would also work. But maybe we can use something like `renderTrigger=` instead? Check the #730

src/components/Dropdown/Dropdown.tsx
```diff
@@ -141,10 +143,12 @@ const DropdownComponent: FC<DropdownProps> = ({
       className={className}
       minWidth={buttonWidth}
     >
-      <TriggerWrapper setButtonWidth={setButtonWidth}>
-        {label}
-        {arrowIcon && <Icon className={theme.arrowIcon} />}
-      </TriggerWrapper>
+      {CustomTrigger ?? (
```
rluders, line 146: What would happen if we don't pass the `CustomTrigger`?

rluders: @paghar in this case, if you don't specify the "CustomTrigger" you don't have any trigger. Right? So, this is not correct.

app/docs/components/dropdown/dropdown.mdx
rluders: @paghar here, what I would prefer, instead passing the `ReactNode` I would prefer to pass a function... using the property `renderTrigger={(theme) => <button></button>}`

The concept is similar to: https://react.dev/reference/react/cloneElement#passing-data-with-a-render-prop

Here is a similar implementation: https://github.com/themesberg/flowbite-react/blob/408aaf3ef34b6b8b1c86f79697d1f9e259647ce1/src/components/Card/Card.spec.tsx#L30

src/components/Dropdown/Dropdown.tsx
paghar: @rluders 
I use the default trigger. When there is no exist
 https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing

src/components/Dropdown/Dropdown.tsx
```diff
@@ -40,6 +40,7 @@ export interface DropdownProps
   inline?: boolean;
   label: ReactNode;
   theme?: DeepPartial<FlowbiteDropdownTheme>;
+  renderTrigger?: () => ReactNode;
```
rluders, line 43: @paghar thank you.

I think that it is almost there. I think that it is important to pass the `theme` as a function parameter, so it can be used to access the theme styles, as, for example... `theme.arrowIcon`.

So probably, I would replace this line with:

```ts
renderTringger?: (theme: FlowbiteDropdownTheme) => ReactNode;
```

Also:

- [ ] Update the documentation example
- [ ] Include some test cases for it

And then, it is good to merge.

rluders (changes requested):

rluders:
@paghar could you rebase it with `main`? Also probably revert the merged with `main`, 'cause it breaks git history. You should always prefer to rebase instead merge. :)

paghar:
@rluders I updated the branch. Please merge it.

rluders:
@paghar Please, revert your merge commit and use rebase instead of merge to update your branch.

rluders:
Closing this one for #834 834

Issue #782, opened 2023-05-31
Add Fathom Analytics to flowbite-react.com


zoltanszogyenyi:
**Is your feature request related to a problem? Please describe.**

Track number of visitors and page visits via an analytics program.

**Describe the solution you'd like**

Use Fathom Analytics as a privacy-friendly solution.

**Describe alternatives you've considered**

Google Analytics (free but not privacy friendly) or Plausible. We already have a subscription for Fathom.

**Additional context**

None.

PR #783, opened 2023-05-31
docs(fathom): add fathom analytics

zoltanszogyenyi:
## Description

Add Fathom Analytics to Flowbite React.

Fixes # [(issue)](https://github.com/themesberg/flowbite-react/issues/782)

## Type of change

## Breaking changes

No breaking changes.

## How Has This Been Tested?

No test required.

## Checklist:

### DIFF
--- a/app/components/fathom.tsx
+++ b/app/components/fathom.tsx
@@ -6,6 +6,7 @@ import { useEffect } from 'react';
 export default function Fathom() {
   const pathname = usePathname();
   const searchParams = useSearchParams();
+
   useEffect(() => {
     load('UXMSXUQI', {
       includedDomains: ['flowbite-react.com'],
--- a/app/components/fathom.tsx
+++ b/app/components/fathom.tsx
@@ -1,0 +1,26 @@
+'use client';
+import { load, trackPageview } from 'fathom-client';
+import { usePathname, useSearchParams } from 'next/navigation';
+import { useEffect } from 'react';
+
+export default function Fathom() {
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
+  useEffect(() => {
+    load('UXMSXUQI', {
+      includedDomains: ['flowbite-react.com'],
+      spa: 'auto',
+    });
+
+    // Initialize Fathom when the app loads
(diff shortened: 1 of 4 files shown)


app/components/fathom.tsx
```diff
@@ -0,0 +1,26 @@
+'use client';
+import { load, trackPageview } from 'fathom-client';
+import { usePathname, useSearchParams } from 'next/navigation';
+import { useEffect } from 'react';
+
+export default function Fathom() {
```
tulup-conner, line 6: If this is being applied to every page, IMO it should just be a const inside of `layout.tsx`, rather than having its own file in `components`. And in this case, it doesn't even render anything - this `useEffect` should just be in `layout.tsx`

app/components/fathom.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const pathname = usePathname();
+  const searchParams = useSearchParams();
```
tulup-conner, line 8: Please add a space after between blocks - e.g. `searchParams` and `useEffect` here

package.json
```diff
@@ -56,6 +56,7 @@
   "dependencies": {
     "@floating-ui/react": "^0.24.1",
     "classnames": "^2.3.2",
+    "fathom-client": "^3.5.0",
```
tulup-conner, line 59: Please move to `devDependencies` and keep in mind that anything you add to `dependencies` becomes bundled with every user's `npm i flowbite-react`, including production deployments for their app which does not include any of our documentation

tulup-conner (changes requested):

app/components/fathom.tsx
zoltanszogyenyi: Can't do that because `layout.tsx` is a server component and using `useEffect` is allowed only for client component. If I make the `layout.tsx` file a client component then exporting meta data is not allowed.

app/components/fathom.tsx
zoltanszogyenyi: Done, not sure why Prettier didn't pick this up.

tulup-conner (approved):

PR #784, opened 2023-05-31
docs: version from package.json

rluders:
## Description

It updates the documentation to display the latest version for `package.json` at the navbar.

## Type of change

- [X] This change contains documentation update

## Breaking changes

None.

## How Has This Been Tested?

I just run it, and check if it was working.

**Test Configuration**:

Not relevant.

## Checklist:

### DIFF
--- a/app/components/navbar.tsx
+++ b/app/components/navbar.tsx
@@ -1,5 +1,5 @@
 import Link from 'next/link';
-import { type FC } from 'react';
+import type { FC } from 'react';
 import { BsGithub } from 'react-icons/bs';
 import { Badge, DarkThemeToggle, Tooltip } from '~/src';
 import pkg from './../../package.json' assert { type: 'json' };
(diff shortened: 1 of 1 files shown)


app/components/navbar.tsx
```diff
@@ -1,7 +1,8 @@
 import Link from 'next/link';
-import type { FC } from 'react';
```
tulup-conner, line 2: This was a mistake I think?

tulup-conner (changes requested):
I think this will work just fine since the `package.json` is referenced from docs only and not the library itself. But check the `import type` for me

app/components/navbar.tsx
rluders: Well, it seems that it is correct. We have it also for other components.
https://typescript-eslint.io/rules/consistent-type-imports/

tulup-conner (approved):

tulup-conner (approved):

PR #840, opened 2023-07-04
feat(dropdown.tsx): Dropdown Accessibility

### LINKED ISSUE #539: Dropdown disabled and loading options
The dropdown feature on the https://flowbite-react.com/dropdown does not provide the loading and disabled options. 

There are already implemented on Buttons category. Thus, it is possible to integrate the option loading={true} (or the div with the spinner) and disabled={true} on the dropdown feature.

### LINKED ISSUE #648: `<Dropdown>` is not accessible by keyboard
**Describe the bug**
`<Dropdown>`s should gain focus, so that the next Tab is the first element in the dropdown.

**To Reproduce**

1. Go to https://flowbite-react.com/dropdown
2. Tab to a Dropdown and hit Enter to expand
3. Notice you cannot tab or use keyboard to navigate dropdown items

### LINKED ISSUE #793: Dropdown dismiss on click only works the first time
**Describe the bug**
Dropdown dismiss on click only works the first time the dropdown menu is opened

**To Reproduce**

1. Trigger drop down
2. Clicking an item dismisses the menu as expected
3. Trigger dropdown a second time
4. Clicking an item does not dismiss the menu

**Expected behavior**
Clicking a dropdown menu item dismisses the menu consistently

**Screenshots**

https://github.com/themesberg/flowbite-react/assets/2526507/28decfc7-ebec-40b5-b93b-5e4b2a71683c

**System information:**

- Device: Mac
- Resolution:
- OS: Monterey
- Browser: Chrome
- Version: 114.0.5735.90

**Project information:**

- Tailwind: 3.0.0
- Flowbite: 1.4.7
- Flowbite React: 0.1.2
- Type: CRA

**Additional context**

nigellima:
## Description

This PR enables the `Dropdown.tsx` component to be navigated by keyboard, improves the Dropdown.Item component to be rendered using `as` prop, and fixes the dismiss on click and `disabled` trigger issues.

Fixes #648. Fixes #793 . Fixes #539 

dropdown-a11y

The package resolution added in `package.json` was necessary due to some errors being thrown in the tests when using `FloatingFocusManager` component. See: https://github.com/focus-trap/focus-trap-react/issues/986

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] This change contains documentation update - PENDING

## How Has This Been Tested?

TBD

## Checklist:

### DIFF
--- a/package.json
+++ b/package.json
@@ -139,6 +139,9 @@
     "react-dom": "^18",
     "tailwindcss": "^3"
   },
+  "resolutions": {
+    "nwsapi": "2.2.2"
+  },
   "private": false,
   "eslintConfig": {
     "extends": [
--- a/src/components/Dropdown/DropdownItem.tsx
+++ b/src/components/Dropdown/DropdownItem.tsx
@@ -10,6 +10,7 @@ import { ButtonBase } from '../Button/ButtonBase';
 import { DropdownContext } from './Dropdown';
 
 export interface FlowbiteDropdownItemTheme {
+  container: string;
   base: string;
   icon: string;
 }
@@ -36,7 +37,7 @@ export const DropdownItem: FC<Props> = ({
   const theme = mergeDeep(useTheme().theme.dropdown.floating.item, customTheme);
 
   return (
-    <li role="menuitem">
+    <li role="menuitem" className={theme.container}>
       <ButtonBase
         ref={ref}
         className={twMerge(theme.base, className)}
--- a/src/components/Dropdown/DropdownItem.tsx
+++ b/src/components/Dropdown/DropdownItem.tsx
@@ -1,24 +1,27 @@
-/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
-/* eslint-disable jsx-a11y/click-events-have-key-events */
-
+import { useListItem } from '@floating-ui/react';
 import type { ComponentProps, FC, PropsWithChildren } from 'react';
+import { useContext } from 'react';
 import { twMerge } from 'tailwind-merge';
 import type { DeepPartial } from '../../';
 import { useTheme } from '../../';
 import { mergeDeep } from '../../helpers/merge-deep';
+import type { ButtonBaseProps } from '../Button/ButtonBase';
+import { ButtonBase } from '../Button/ButtonBase';
+import { DropdownContext } from './Dropdown';
 
 export interface FlowbiteDropdownItemTheme {
   base: string;
   icon: string;
 }
 
-export interface DropdownItemProps extends PropsWithChildren, ComponentProps<'li'> {
+export interface DropdownItemProps extends PropsWithChildren, ButtonBaseProps {
   icon?: FC<ComponentProps<'svg'>>;
   onClick?: () => void;
   theme?: DeepPartial<FlowbiteDropdownItemTheme>;
 }
+interface Props extends DropdownItemProps, Record<string, unknown> {}
 
-export const DropdownItem: FC<DropdownItemProps> = ({
+export const DropdownItem: FC<Props> = ({
   children,
   className,
   icon: Icon,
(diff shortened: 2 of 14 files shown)


src/helpers/floating.ts
```diff
@@ -37,3 +39,25 @@ export const getArrowPlacement = ({ placement }: { placement: Placement }): Plac
     left: 'right',
   }[placement.split('-')[0]] as Placement;
 };
+
+export const mergeWrapperClassName = ({
```
rluders, line 43: You created this one because it is used in two places, right? I'm wondering if this case isn't worth repeating the code. But not against leaving it as a helper. I'm just thinking loud.

package.json
```diff
@@ -139,6 +139,9 @@
     "react-dom": "^18",
     "tailwindcss": "^3"
   },
+  "resolutions": {
+    "nwsapi": "2.2.2"
```
rluders, line 143: Do we need this? Why?

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -1,37 +1,57 @@
...
+  const { activeIndex, dismissOnClick, getItemProps, handleSelect } = useContext(DropdownContext);
+  const isActive = activeIndex === index;
+
   const theme = mergeDeep(useTheme().theme.dropdown.floating.item, customTheme);
 
   return (
-    <li className={twMerge(theme.base, className)} onClick={onClick} {...props}>
-      {Icon && <Icon className={theme.icon} />}
-      {children}
+    <li role="menuitem">
```
rluders, line 39: Where is the theme for `li` here?

src/components/Dropdown/DropdownItem.tsx
```diff
@@ -1,37 +1,57 @@
...
+      <ButtonBase
+        ref={ref}
+        className={twMerge(theme.base, className)}
```
rluders, line 42: I think that this is not the `theme.base`, maybe this is something else...

rluders (changes requested):
Wow! Really nice, @nigellima. I added just a few comments, but in general, it looks great.

package.json
rluders: Oh, I see why... Nevermind.

src/helpers/floating.ts
nigellima: @rluders I actually didnt like this utility function either. I will repeat the twMerge in both places then

src/components/Dropdown/DropdownItem.tsx
nigellima: @rluders it actually is. I left the `li menuitem` more of as a container to hold the accessible role, then the user will be able to style the `ButtonBase` as they like or to change the base tag - and still have the container item as role menuitem

nigellima: Yeah, makes sense heheh I will add it then

rluders (approved):
Thank you, @nigellima.

tulup-conner:
This is another fantastic PR, thank you very much!

maxlibin:
In light of the changes made in this link: https://github.com/themesberg/flowbite-react/pull/840/files#diff-32fa9ddaf531f010c646834b9023bf97df1cd80d6de82762a56bc99514594f83R238, it appears that everything within the dropdown is now contained in an <ul> tag. Previously, I could incorporate a form complete with input fields inside the dropdown; however, that functionality no longer seems operational. Is this an intentional modification?

Additionally, I found some documentation here: https://github.com/themesberg/flowbite-react/pull/840/files#diff-d5c7e7428127c9af056fda02409bfef933507c13fa5dad102ee8dc33415b3c55R94. It suggests that I can use as="span". However, this suggestion raises a question for me: Are we now inserting a span element within the ul? To me, this implementation feels unusual.

Please provide clarification if my understanding of these changes is mistaken. I appreciate your guidance 🙏.

rluders:
The `Dropdown` children will be always a `Dropdown.Item`, the component that you can "morph" is the `Trigger`, which is outside the `ul` and the `Dropdown.Item`'s button that is inside the `li`. So... Unless I'm not seeing something here, I don't see any issue with this implementation. And yes, the `ul > li` structure was desired.

maxlibin:
thanks for clarification

PR #868, opened 2023-07-19
fix(button.tsx): Relative Spinner size

### LINKED ISSUE #850: Spinner icon inside Button does not adjust to button size
**Describe the bug**
When the Button component is smaller (like `sm` and `xs`) the `Spinner` inside it makes the component increase in size when `isProcessing={true}`

button-spinner

**To Reproduce**

1. Add `xs` prop to Button component
2. add isProcessing={true}
3. You should see the component getting taller

**Expected behavior**
The Button component should smoothly increase in width to fit the Spinner and its size should be relative to the Button size.

The animation of fitting the Spinner is more of a nice to have (example below). But the relative size I believe it's a bug.

button-loading-ideal

**Project information:**

- Flowbite React: 0.4.11

nigellima:

*Summarize the changes made and the motivation behind them.*

This PR makes the <Spinner /> component inside the <Button /> component to be relative to the Button size when `isProcessing=true`. It also adds an width animation when the flag is toggled.

processing

Fix #850 

### DIFF
--- a/app/docs/components/button/button.mdx
+++ b/app/docs/components/button/button.mdx
@@ -2,6 +2,7 @@ import { CodePreview } from '~/app/components/code-preview';
 import { Button, theme } from '~/src';
 import { HiOutlineArrowRight, HiShoppingCart } from 'react-icons/hi';
 import Link from 'next/link';
+import { AiOutlineLoading } from 'react-icons/ai';
 
 ## Table of Contents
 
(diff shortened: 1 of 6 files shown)


nigellima:
@rluders weird, the test in the CI / build workflow does not fail locally for me.

tulup-conner:
https://github.com/themesberg/flowbite-react/actions/runs/5605163775/jobs/10253955428#step:8:342 I'm also not sure why exactly this is happening

rluders:
I notice that some tests are failing, not 'cause the test subject is failing but because the test itself has some implementation issue... maybe some rendering, or delay process that is causing the issue. Probable we should create an issue to fix it, I'm sure that the problem isn't related to this PR specifically.

Issue #823, opened 2023-06-22
Alert


rluders:

batorlancs:
I have been using this library and I am looking to get my first issue! Could I get assigned to this problem?

PR #874, opened 2023-07-22
test(alert): css classes verification

benilyxdd:

Added tests to verify if the CSS classes are aligned with the standard of Flowbite

#823

### DIFF
--- a/src/components/Alert/Alert.spec.tsx
+++ b/src/components/Alert/Alert.spec.tsx
@@ -153,6 +153,54 @@ describe.concurrent('Components / Alert', () => {
       expect(onDismiss).toHaveBeenCalled();
     });
   });
+
+  // https://flowbite.com/docs/components/alerts/#default-alert
+  describe('Classname', async () => {
+    it('should have info classes', () => {
+      render(<Alert color="info" />);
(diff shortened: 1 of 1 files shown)


tulup-conner:
Thanks for your contribution! I like the idea of this, but I would prefer to find a solution that scales to every component without writing hundreds of tests. I'm not really sure what that is yet - but I think this would be a great place to experiment with that.

benilyxdd:
@tulup-conner 

> every component without writing hundreds of tests
Do you mean the tests should also test the other components like Accordion and Button instead of just Alert?

rluders:
> Thanks for your contribution! I like the idea of this, but I would prefer to find a solution that scales to every component without writing hundreds of tests. I'm not really sure what that is yet - but I think this would be a great place to experiment with that.

This is going to be hard, TBH. I don't see any other way to do it automatically or without writing a lot of unit tests. Probably in this case simple approach, even than not perfect, would be better than nothing.

tulup-conner:
Thanks for contributing, but I don't think we have any plans to implement something like this because it would be such a monumental task. Definitely check out one of the other open issues if you have some time and want to help out, it would be greatly appreciated :D

PR #879, opened 2023-07-24
docs(carbon ads): integrate Carbon Ads

sayzlim:
Change docs-content-layout to client component.

Integrate Carbon Ads into the documentation and homepage. There is a change to the `docs-content-layout` to client component because we need `useEffect` to inject Carbon script into the component where the ad should be visible. Here are the screenshots of how the ad looks like:

Homepage

Docs

### DIFF
--- a/app/components/carbon-ads.tsx
+++ b/app/components/carbon-ads.tsx
@@ -3,16 +3,20 @@
 import { useEffect } from 'react';
 
 const CarbonAds = () => {
-	useEffect(() => {
-		const script = document.createElement('script');
-		script.src = '//cdn.carbonads.com/carbon.js?serve=CEAIC53L&placement=flowbite-reactcom';
-		script.id = '_carbonads_js';
-		script.async = true;
+  useEffect(() => {
+    const script = document.createElement('script');
+    script.src = '//cdn.carbonads.com/carbon.js?serve=CEAIC53L&placement=flowbite-reactcom';
+    script.id = '_carbonads_js';
+    script.async = true;
 
-		document.querySelectorAll('#carbon-container')[0].appendChild(script);
-	}, []);
+    document.querySelectorAll('#carbon-container')[0].appendChild(script);
+  }, []);
 
-	return <div id="carbon-container"></div>;
+  return (
+    <div className="fixed bottom-5 right-5 z-50 hidden sm:block">
+      <div id="carbon-container"></div>
+    </div>
+  );
 };
 
 export default CarbonAds;
--- a/app/components/carbon-ads.tsx
+++ b/app/components/carbon-ads.tsx
@@ -1,0 +1,18 @@
+'use client';
+
+import { useEffect } from 'react';
+
+const CarbonAds = () => {
+	useEffect(() => {
+		const script = document.createElement('script');
+		script.src = '//cdn.carbonads.com/carbon.js?serve=CEAIC53L&placement=flowbite-reactcom';
+		script.id = '_carbonads_js';
+		script.async = true;
+
+		document.querySelectorAll('#carbon-container')[0].appendChild(script);
+	}, []);
+
+	return <div id="carbon-container"></div>;
+};
+
+export default CarbonAds;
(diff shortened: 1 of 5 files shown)


zoltanszogyenyi:
Hey @sayzlim,

Thanks for helping with the Carbon Ads integration.

First of all, I've moved the ad to the bottom right side of the page similar to how we have it for [flowbite.com](https://flowbite.com).

There's an issue that it seems like it's duplicated after rendering, there are also some errors in the console which I think are happening because of the bad rendering of the ads.

Here's how it looks for me:

Screenshot 2023-07-24 at 12 17 38

Cheers,
Zoltan

zoltanszogyenyi:
Also, not sure it's a good idea to change it to a client component as it introduced a lot of errors in the console:

Screenshot 2023-07-24 at 12 23 27

Need your opinion on this @tulup-conner too.

sayzlim:
@zoltanszogyenyi duplicate ad is expected if you run in dev mode due to how React runs every component twice to catch issues or errors. If you run `npm run build` and test the built site with `npm run start`, the ad should only appear once per page. I don't see any errors when checking the built site.

I will defer to @tulup-conner for the best practice here. Ideally we keep the ad in the documentation (better mobile visibility too) so we don't have to set fixed positioning for the ad, but it's also okay if you don't mind having the fixed ad on the docs.

zoltanszogyenyi:
To be fixed:

- [ ] the ad should re-render as you navigate through pages to correctly resemble impressions and revenue
- [ ] fix the duplicate render visually
- [ ] keep the styles
- [ ] no new errors introduced (I removed the `use client` which introduced most of them)

zoltanszogyenyi:
> @zoltanszogyenyi duplicate ad is expected if you run in dev mode due to how React runs every component twice to catch issues or errors. If you run `npm run build` and test the built site with `npm run start`, the ad should only appear once per page. I don't see any errors when checking the built site.
> 
> I will defer to @tulup-conner for the best practice here. Ideally we keep the ad in the documentation (better mobile visibility too) so we don't have to set fixed positioning for the ad, but it's also okay if you don't mind having the fixed ad on the docs.

Hey, Lim!

Mobile views for Flowbite React based on our analytics is below 5% - but we can still show it on mobile devices. I prefer to show it in a fixed state on the bottom right side and we'll make updates if needed in the future.

@tulup-conner is the one to review the PR in terms of technical review, however, I would appreciate if the ad would re-render and generate a new impression as you navigate throughout the pages. Even if the original duplicate ad is only a "dev" problem.

Cheers,
Zoltan

sayzlim:
@zoltanszogyenyi I can confirm in the built version that the ad re-render whenever the page page changes. Check out the [recorded video of the local built page](https://share.cleanshot.com/NrtmHqx3). This will ensure all the possible impressions are recorded by our ad server.

zoltanszogyenyi:
@sayzlim gotcha - if the checklist above is fixed after build and @tulup-conner approves the review then it's good to go!

tulup-conner:
Hey @sayzlim thank you for adding this! I made the following changes:

1. Disable Carbon Ads in development mode; it doesn't help to track users running the dev server in the first place, and also causes the ads to duplicate, which makes development more challenging
2. Converted `next/router` `useRouter()` to `usePathname()` as required by Next.js 13
3. Created a custom `Window` wrapper in TypeScript that includes the magic `_carbonads` handle so Next.js builds succeed

It's a bit unfortunate that the Carbon JavaScript doesn't work in the intuitive React way - what I would like to do is just insert the `<script>` as JSX inside `#carbon-container`, but it doesn't seem to work without the traditional DOM manipulation you have provided. That isn't important at all, though, so, oh well!

I can confirm that ads don't show up in dev mode, and they do not repeat/stack in production! I think this is ready to go so I will merge now, which means we will start displaying ads effectively immediately on flowbite-react.com.

tulup-conner:
Also, you didn't create a feature branch in this PR, so this is going to be implemented in a separate Pull Request.

Issue #855, opened 2023-07-13
Different heights of buttons in Button.Group when text is rendered in more than one line due to lack of space


victorfunes:
- [ Yes ] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [ Yes ] I have tested the latest version

## Steps to reproduce

1. Define a Button.group with different number and length of words and shrink the space available for all the component

## Current behavior

Now each button has a different height depending on the words of each button:

<img width="264" alt="image" src="https://github.com/themesberg/flowbite-react/assets/108006690/79d676f2-c0a9-431d-9939-a1f39c0a482a">

## Expected behavior

All the buttons should have the same height.

## Context

I was checking responsive design for mobile devices, and I have seen this strange behaviour.

```
        <Button.Group>
          <Button size="sm" color="light" onPointerDown={filterLastYear}>
            <span className={internalFilterDate === FilterDate.LAST_YEAR ? selectedButtonClassname : undefined}>
              Last year
            </span>
          </Button>
          <Button size="sm" color="light" onPointerDown={filterLastMonth}>
            <span className={internalFilterDate === FilterDate.LAST_MONTH ? selectedButtonClassname : undefined}>
              Last 30 days
            </span>
          </Button>
          <Button size="sm" color="light" onPointerDown={filterLastWeek}>
            <span className={internalFilterDate === FilterDate.LAST_WEEK ? selectedButtonClassname : undefined}>
              Last 7 days
            </span>
          </Button>
          <Button size="sm" color="light" onPointerDown={filterAlways}>
            <span className={internalFilterDate === FilterDate.ALWAYS ? selectedButtonClassname : undefined}>
              Always
            </span>
          </Button>
        </Button.Group>
```

000000armor:
@rluders Hi! I opened a PR with a fix. Please check it when it's possible.

SutuSebastian:
Fixed by #1498  in [flowbite-react@0.11.0](https://github.com/themesberg/flowbite-react/releases/tag/flowbite-react%400.11.0) release.

Please see [Docs - Colors - Customizing Colors](https://flowbite-react.com/docs/customize/colors#customizing-colors).

PR #906, opened 2023-08-07
fix(button): allow height of buttons stretch

000000armor:

Add styles to parent container to allow children stretch

#855


rluders:
Thanks, @000000armor.

I think that is worth to test what would happened if we have something like this:

``` jsx
<Button.Group>
      <Button color="gray">
        <HiUserCircle className="mr-3 h-4 w-4" />
        <p>
          Profile
        </p>
      </Button>
      <Button color="gray">
        <HiAdjustments className="mr-3 h-4 w-4" />
        <p>
          Settings <br/> with breakline
        </p>
      </Button>
      <Button color="gray">
        <HiCloudDownload className="mr-3 h-4 w-4" />
        <p>
          Messages
        </p>
      </Button>
    </Button.Group>
```

I think that the icon and text will not be centered. Can you check?

000000armor:
@rluders yes, I'll check it. that's a nice corner case. Maybe we should add it to storybook example to show how the component should behave in certain situations?

000000armor:
Screenshot 2023-08-08 at 16 08 10
@rluders is it expected behaviour for provided case?

Issue #925, opened 2023-08-22
ToggleSwitch simplified


victorfunes:
- [X] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [X] I have tested the latest version

## Summary

The ToggleSwitch component is requiring always parameters that are not always being to be used.
Request: change to optional some of them.

## Context

I am creating a form just to visualize information, and due to the screen design, I cannot use the embedded label, but it is mandatory to define one.
As it is only for information, it is not going to change, but the onChange function is also mandatory.

Example: I have to implement this design using ```label=""``` and ```onChange={() => {}}``` and generating unnecessary warnings in my code.

![image](https://github.com/themesberg/flowbite-react/assets/108006690/e221d0df-3ff8-4b59-97d8-8273a4bec5ff)

```
            <Label
              htmlFor={"automatedAssignment"}
              className="text-xs font-normal text-gray-900 select-none">
              Allow automatic assignment
            </Label>
            <ToggleSwitch
              id="automatedAssignment"
              label=""
              color="blue"
              checked={automaticAssignmentEnabled ?? false}
              onChange={() => {}}
            />
```

andreaTarascioSophany:
Hi! I'm picking up this issue

rluders:
Still missing the `onChange`

SutuSebastian:
> Still missing the `onChange`

Unless we make it have internal state (or have the `button` listen to vanilla `input` checked state) so that it becomes uncontrolled, we need to enforce `checked` and `onChange` props.

PR #928, opened 2023-08-23
refactor(toggleswitch.tsx): toggleSwitch optional props

andreaTarascioSophany:
ToggleSwitch optional props

#925 


tulup-conner (changes requested):
Thank you for taking this issue on! I know we have tons of existing code where we do this, but we would like to change the pattern of

```
{booleanCondition && <SomeJSX />}
```

to

```
{booleanCondition ? <SomeJSX /> : null}
```

The first example will cause a 0 to appear if the condition is false.

Also, if you could add an example to the [docs for `<ToggleSwitch>`](https://www.flowbite-react.com/docs/components/forms#toggle-switch) where the `label` is omitted, that would be greatly appreciated.

andreaTarascioSophany:
> 

Okay! I'll make the changes asap

rluders (approved):

Issue #924, opened 2023-08-22
Add Option To Style Tab Panel Container in Theme


ogheneovo12:
- [ ] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [ ] I have tested the latest version

## Summary
I couldn't find any theme option to style the container div of the tab panels, i needed this to be able to achieve vertical tabs

Describe how it should work, and provide examples of the solution, which might include screenshots or code snippets.
the theme should have an option to style the tab panel container
## Context

What are you trying to accomplish? How is your use case affected by not having this feature?
i am trying to achieve vertical tabs, so a applied flex-row to the tab wrapper, i noticed the tab panels weren't full width, they had a container div that wrapped, i needed a way to specify flex-grow on the tab panel container but was unable to find such option in the theme. see screenshot below

Tab Panel Container Wrapper by default
![Screenshot (1054)](https://github.com/themesberg/flowbite-react/assets/49083654/989aa8af-f40b-4573-87c9-2086dd0257e7)

applying flex grow manually via dev tools
![Screenshot (1055)](https://github.com/themesberg/flowbite-react/assets/49083654/34a43abf-99c1-4e25-87a5-66363d716718)

PR #929, opened 2023-08-23
fix(tabs): added tabs container div style and theme

DavideFrancescon:

Tabs inside Tab component are wrapped in an unstyled div, and in no way this was possible to add any style to it. I added a now theme key "tabsitemcontainer" to let the div take in style. As it was possible to style the divs inside (the single tabs), it was kinda stupid not being able to add style to a div that contains them, as it brought issues with dimensions.

#924

Yes. I added a key to the theme of the tabs, the theme i added is completely empty and is done only for users to implement some custom styles if necessary.
```
    tabitemcontainer: {
      base: '',
      styles: {
        default: '',
        underline: '',
        pills: '',
        fullWidth: '',
      },
    }
```
Interface:
```
{
  base: string;
  tablist: {
    base: string;
    styles: TabStyles;
    tabitemcontainer: {
      base: string;
      styles: TabStyles;
    };
    tabitem: {
      base: string;
      styles: TabStyleItem<TabStyles>;
      icon: string;
    };
  };
  tabpanel: string;
}
```

### DIFF
--- a/src/components/Tab/Tabs.tsx
+++ b/src/components/Tab/Tabs.tsx
@@ -152,7 +152,7 @@ export const TabsComponent = forwardRef<TabsRef, TabsProps>(
             </button>
           ))}
         </div>
-        <div className={twMerge(theme.tablist.tabitemcontainer.base, tabItemContainerStyle)}>
+        <div className={twMerge(theme.tabitemcontainer.base, tabItemContainerStyle)}>
           {tabs.map((tab, index) => (
             <div
               key={index}
--- a/src/components/Tab/Tabs.tsx
+++ b/src/components/Tab/Tabs.tsx
@@ -152,7 +152,7 @@ export const TabsComponent = forwardRef<TabsRef, TabsProps>(
             </button>
           ))}
         </div>
-        <div className={twMerge(tabItemContainerStyle)}>
+        <div className={twMerge(theme.tablist.tabitemcontainer.base, tabItemContainerStyle)}>
           {tabs.map((tab, index) => (
             <div
               key={index}
--- a/src/components/Tab/Tabs.tsx
+++ b/src/components/Tab/Tabs.tsx
@@ -12,6 +12,10 @@ export interface FlowbiteTabTheme {
   tablist: {
     base: string;
     styles: TabStyles;
+    tabitemcontainer: {
+      base: string;
+      styles: TabStyles;
+    };
     tabitem: {
       base: string;
       styles: TabStyleItem<TabStyles>;
@@ -104,6 +108,7 @@ export const TabsComponent = forwardRef<TabsRef, TabsProps>(
     };
 
     const tabItemStyle = theme.tablist.tabitem.styles[style];
+    const tabItemContainerStyle = theme.tablist.tabitemcontainer.styles[style];
 
     useEffect(() => {
       tabRefs.current[focusedTab]?.focus();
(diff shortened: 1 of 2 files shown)


src/components/Tab/Tabs.tsx
```diff
@@ -12,6 +12,10 @@ export interface FlowbiteTabTheme {
   tablist: {
     base: string;
     styles: TabStyles;
+    tabitemcontainer: {
```
rluders, line 15: I would say that probably we want to rename this one to `tabItemContainer`.

src/components/Tab/Tabs.tsx
```diff
@@ -146,7 +151,7 @@ export const TabsComponent = forwardRef<TabsRef, TabsProps>(
             </button>
           ))}
         </div>
-        <div>
+        <div className={twMerge(tabItemContainerStyle)}>
```
rluders, line 154: I think that you want to also merge this style with the `tabItemContainer.base`, right?

rluders (changes requested):

src/components/Tab/Tabs.tsx
DavideFrancescon: I followed the naming of the other components, they are respectively called: tabitem and tablists. So having only this one as camel case is probably going to confuse everyone. We could change every key to a camel case one, but it probably need a separate PR.

src/components/Tab/Tabs.tsx
DavideFrancescon: Oh yea right, thanks!

DavideFrancescon:
@rluders I added a new commit with another bug i noticed while fixing the issue. Let me know if it's okay to leave it here, maybe we can change the title/description a bit

ogheneovo12:
it kind of feels weird that the specification is under tabLists but in actual reality it's more  related to tabPanel, it's the container that wraps the tab panels,

```
<div role="tablist"></div>
<div role="assumedPanelContainerRole">
       <div role="panel"></div>
      <div role="panel"></div>
</div>
```
i wouldn't want the tabPanel structure to be changed for backward compatibility reasons

probably just a tabPanelContainer property at the root of the tab definition

```
{
  base: string;
  tablist: {
    base: string;
    styles: TabStyles;
    tabitem: {
      base: string;
      styles: TabStyleItem<TabStyles>;
      icon: string;
    };
  };
  tabpanel: string;
  tabPanelContainer:"" //or a more verbose options based on tab style
}
```

DavideFrancescon:
The thing is, all the panels are inside the list, so I guess I can refactor all the theme for that component to have it make more sense (breaking all backward compatibility), or just take that one out leaving the rest inside (this is probably what I'll do).

 Other than this, do i keep the same all lower case naming or do I change only that one?

DavideFrancescon:
While testing a bit I noticed another bug with the focus ring, I changed some code to fix it. I'll post it here before commiting, i want to know if it's production ready like this or not

How I tried to fix it yesterday:
```
style={{zIndex: index === focusedTab ? 2 : 1 }}
```

New code to fix it(this works fine:
```         
onFocusCapture={() => {
    tabRefs.current[index].style.zIndex = '2';
}}
onBlur={() => {
    tabRefs.current[index].style.zIndex = '0';
}}
```

The think i'm not sure about is if those function getting called like that are going to influence something. I don't think they will but i want a second opinion on it. Thanks to whoever responds!

rluders (approved):

PR #941, opened 2023-08-29
fix/dropdown-input

### LINKED ISSUE #916: Input inside a dropdown is not working (can't type)

## Steps to reproduce
Code given bellow.

## Current behavior

When I put a input inside the Dropdown header I can't type in that input box. But if I put it outside I can type. I can only increase using the up / down arrow in the right side of the input but I want to be able to type also.

## Expected behavior

I should be able to type there. not only up/down arrow.

## Context
So I have a dropdown and people can select area as numbers.
image

```jsx
  <Dropdown className={styles.dropdown} label={'Area'}>
              <Dropdown.Header>
                <label htmlFor="purpose" className="block text-sm font-medium mb-2">
                  Area (sqft)
                </label>
                <div>
                  <div className="mb-2 block">
                    <Label htmlFor="minimum-area" value="Minimum sqft" />
                  </div>
                  <TextInput
                    id="minimum-area"
                    placeholder="Minimum sqft"
                    onChange={(e: any) => {
                      setMinArea(e.target.value as number)
                      console.log(e.target.value)
                    }}
                    value={minArea}
                    sizing="sm"
                    step={100}
                    min={0}
                    type="number"
                  />
                </div>
                <div>
                  <div className="mb-2 block">
                    <Label htmlFor="maximum-area" value="Maximum sqft" />
                  </div>
                  <TextInput
                    id="maximum-area"
                    onChange={(e: any) => setMaxArea(e.target.value as number)}
                    placeholder="Maximum sqft"
                    value={maxArea}
                    sizing="sm"
                    step={100}
                    min={0}
                    type="number"
                  />
                </div>
              </Dropdown.Header>
              <Dropdown.Item className="hover:bg-transparent">
                <div className="flex justify-between w-[100%] ">
                  <Button
                    onClick={(e: any) => {
                      setMinArea(0)
                      setMaxArea(500)
                    }}
                    className="w-[46%]"
                  >
                    Reset
                  </Button>
                  <Button className="w-[46%]">Done</Button>
                </div>
              </Dropdown.Item>
            </Dropdown>
```

ChrisCoastal:

This pull request removes Floating-UI's `useTypeahead` from the `Dropdown`. The typeahead functionality is causing unexpected behaviour for dropdowns containing `<input>` elements or other UI expecting user keyboard events, as the feature stops the keydown event from propagating.

fix #916

There are no breaking changes.

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -1,5 +1,5 @@
 import type { ExtendedRefs, useInteractions } from '@floating-ui/react';
-import { FloatingFocusManager, FloatingList, useListNavigation, useTypeahead } from '@floating-ui/react';
+import { FloatingFocusManager, FloatingList, useListNavigation } from '@floating-ui/react';
 import type {
   ComponentProps,
   Dispatch,
@@ -155,17 +155,6 @@ const DropdownComponent: FC<DropdownProps> = ({
     setOpen(false);
   }, []);
 
-  const handleTypeaheadMatch = useCallback(
-    (index: number | null) => {
-      if (open) {
-        setActiveIndex(index);
-      } else {
-        handleSelect(index);
-      }
-    },
-    [open, handleSelect],
-  );
-
   const { context, floatingStyles, refs } = useBaseFloating<HTMLButtonElement>({
     open,
     setOpen,
@@ -179,18 +168,18 @@ const DropdownComponent: FC<DropdownProps> = ({
     onNavigate: setActiveIndex,
   });
 
-  const typeahead = useTypeahead(context, {
-    listRef: labelsRef,
-    activeIndex,
-    selectedIndex,
-    onMatch: handleTypeaheadMatch,
-  });
+  // const typeahead = useTypeahead(context, {
+  //   listRef: labelsRef,
+  //   activeIndex,
+  //   selectedIndex,
+  //   onMatch: handleTypeaheadMatch,
+  // });
 
   const { getReferenceProps, getFloatingProps, getItemProps } = useFloatingInteractions({
     context,
     role: 'menu',
     trigger,
-    interactions: [listNav, typeahead],
+    interactions: [listNav],
   });
 
   const Icon = useMemo(() => {
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -26,7 +26,7 @@ import type { FlowbiteDropdownItemTheme } from './DropdownItem';
 import { DropdownItem } from './DropdownItem';
 
 import { twMerge } from 'tailwind-merge';
-import { useBaseFLoating, useFloatingInteractions } from '../../helpers/use-floating';
+import { useBaseFloating, useFloatingInteractions } from '../../helpers/use-floating';
 
 export interface FlowbiteDropdownFloatingTheme
   extends FlowbiteFloatingTheme,
@@ -166,7 +166,7 @@ const DropdownComponent: FC<DropdownProps> = ({
     [open, handleSelect],
   );
 
-  const { context, floatingStyles, refs } = useBaseFLoating<HTMLButtonElement>({
+  const { context, floatingStyles, refs } = useBaseFloating<HTMLButtonElement>({
     open,
     setOpen,
     placement,
--- a/src/components/Dropdown/Dropdown.spec.tsx
+++ b/src/components/Dropdown/Dropdown.spec.tsx
@@ -5,8 +5,6 @@ import { describe, expect, it } from 'vitest';
 import type { DropdownProps } from './Dropdown';
 import { Dropdown } from './Dropdown';
 
-const delay = async (delayTime: number) => await new Promise((r) => setTimeout(r, delayTime));
-
 describe('Components / Dropdown', () => {
   describe('A11y', async () => {
     it('should use `role="menu"` in menu container', async () => {
@@ -74,20 +72,6 @@ describe('Components / Dropdown', () => {
 
       await act(() => fireEvent.keyDown(button(), { key: 'ArrowDown', code: 'ArrowDown' }));
       expect(dropdown()).toBeInTheDocument();
-    });
-
-    it('should focus matching item when user types the first option char and dropdown is open', async () => {
-      const user = userEvent.setup();
-      render(<TestDropdown />);
-
-      await act(() => user.click(button()));
-      expect(dropdown()).toBeInTheDocument();
-
-      await act(() => fireEvent.keyDown(button(), { key: 'S', code: 'KeyS' }));
-      await delay(20);
-
-      const item = screen.getByText('Settings');
-      expect(item).toHaveFocus();
     });
   });
 
--- a/src/components/Floating/Floating.tsx
+++ b/src/components/Floating/Floating.tsx
@@ -4,7 +4,7 @@ import type { ComponentProps, FC, PropsWithChildren, ReactNode } from 'react';
 import { useEffect, useRef, useState } from 'react';
 import { twMerge } from 'tailwind-merge';
 import { getArrowPlacement } from '../../helpers/floating';
-import { useBaseFLoating, useFloatingInteractions } from '../../helpers/use-floating';
+import { useBaseFloating, useFloatingInteractions } from '../../helpers/use-floating';
 
 export interface FlowbiteFloatingTheme {
   arrow: FlowbiteFloatingArrowTheme;
@@ -62,7 +62,7 @@ export const Floating: FC<FloatingProps> = ({
   const arrowRef = useRef<HTMLDivElement>(null);
   const [open, setOpen] = useState(false);
 
-  const floatingProperties = useBaseFLoating({
+  const floatingProperties = useBaseFloating({
     open,
     placement,
     arrowRef,
--- a/src/helpers/use-floating.ts
+++ b/src/helpers/use-floating.ts
@@ -21,7 +21,7 @@ export type UseBaseFloatingParams = {
   setOpen: Dispatch<SetStateAction<boolean>>;
 };
 
-export const useBaseFLoating = <Type extends ReferenceType>({
+export const useBaseFloating = <Type extends ReferenceType>({
   open,
   arrowRef,
   placement = 'top',


src/components/Dropdown/Dropdown.tsx
```diff
@@ -179,18 +168,18 @@ const DropdownComponent: FC<DropdownProps> = ({
     onNavigate: setActiveIndex,
   });
 
-  const typeahead = useTypeahead(context, {
-    listRef: labelsRef,
-    activeIndex,
-    selectedIndex,
-    onMatch: handleTypeaheadMatch,
-  });
+  // const typeahead = useTypeahead(context, {
```
nigellima, line 171: Should not this be removed? Or maybe have a comment above it describing why this needs to be commented but kept

rluders:
Thanks, @ChrisCoastal. I think that aside @nigellima review, it looks good.

I have a suggestion, and this is not a requirement: I'm wondering if we can't add an example of a dropdown with Input to the documentation. I'm trying to think of ideas to expand the documentation examples, creating more practical examples. WDYT?

nigellima:
Wondering here if we are not mixing up the concepts of a Popover with Dropdowns.

Popovers make sense to have forms, inputs or anything inside it. Which is not the same case for Dropdowns.

For example, when I see this (from the referenced issue) to me it looks like a Popover component, not a Dropdown:

image

rluders:
It is possible, @nigellima...  After my comment about the document example, I was looking at flowbite documentation, searching for a good example to be documented... and surprise:

https://flowbite.com/docs/components/popover/

So, my two cents here... We may not want to "fix" the Dropdown, however, what we would like to have is to... introduce the Popover component. Does it make sense?

nigellima:
> It is possible, @nigellima... After my comment about the document example, I was looking at flowbite documentation, searching for a good example to be documented... and surprise:
> 
> https://flowbite.com/docs/components/popover/
> 
> So, my two cents here... We may not want to "fix" the Dropdown, however, what we would like to have is to... introduce the Popover component. Does it make sense?

Yes, I completely agree. Type-head is one of the essential features for Dropdown A11y, I believe we shoundlt be removing it.

There is a Dropdown example in the core library which has a search input in the `Dropdown.Header`:
image

This is a fair use-case TBH. So IMO I believe the ideal solution here for the bug reported is to either disable the Dropdown typehead if an input is found within the Dropdown component or find a way to exclude the `Dropdown.Header` from the type-head feature if it's focused (not sure if possible with floating-ui).

And yes, the introduction of `Popover` with some appealing examples would be perfect and also drive away users to use Dropdown for this kind of use-cases

rluders:
All right, agreed. So...

- [ ] Type-head shouldn't be removed;
- [ ] `Dropdown.Header` needs to not be affected by type-head;
- [ ] `Dropdown.Footer` needs to be introduced;
- [ ] Introduce the *Dropdown with Search* example to the docs;

nigellima:
> Thanks for all the feedback. That seems like a great solution for keeping the typeahead functionality, which is ideal.
> 
> > * [ ]   Type-head shouldn't be removed;
> > * [ ]   `Dropdown.Header` needs to not be affected by type-head;
> > * [ ]   `Dropdown.Footer` needs to be introduced;
> > * [ ]   Introduce the _Dropdown with Search_ example to the docs;
> 
> I'll have some time to add these items this morning and update the pull request.

For the example of the Dropdown with search you see that there are checkboxes there. These checkboxes should be treated differently too... so, maybe you could introduce new child components like `Dropdown.Checkbox` and `Dropdown.Radio` which are not affected by the typehead too.

Basically typehead should work only on `Dropdown.Item` child component

rluders:
> Hello, this is my first time messaging here. Over the last week I've been doing a spike to investigate using this library as a base for an upcoming design system project.
> 
> I've come across this limitation too and wanted to share the design I'm currently trying to implement.
> 
> image
> The required component is a nested select with an outter search functionality. When a user selects an option it is added to an external list.
> 
> Currently the typehead also blocks outter inputs from being used while the drop-down options are open. Just wanted to share this in case it might change an approach you might take with this.
> 
> Completely unrelated but the dropdown also lacks a onOpen/closed callback and Dropdown.Items do not have a disabled prop. But I'll create a separate ticket for those issues during the week.

AFAIK, there is no component like this for Flowbite, and also, it doesn't actually seem to be a dropdown.

ChrisCoastal:
TL;DR
After looking at this and Floating UI more in depth, my feeling now is:

- `UseTypeahead` should be removed or opt-in via props on the Dropdown
- if `UseTypeahead` is kept opt-in, it will need a way to be toggled off if there is to be any sort of input is allowed in a dropdown (adding `Header` or `Footer` will not fix the issues, and might create more navigation issues with `UseListNavigation`).
- an idea for toggling off `UseTypeahead` would be to check if an input is currently focussed (something like document.activeElement?.tag === 'INPUT') but I would be interested if there were a suggestion for a better way to do a toggle
- `UseListNavigation` should possibly also be opt-in as it always overrides (does not allow) native tab navigation; depending on the rest of the UI, this may not always be consistent/intuitive
- if the `Dropdown` is allowing focus to leave (as it currently does with `modal={false}` prop, there will be the issue that @Aaron-Conroy ran into with `UseTypeahead` blocking inputs elsewhere in the UI

So, my proposal would be (obviously open to suggestions):

- [ ] remove `UseTypeahead`
- [ ] make `UseListNavigation` opt-in via props
- [ ] consider changing how `Dropdown.Item` is composed (maybe passing props rather than allowing children to put some guardrails on ways the component cannot be used)
- [ ] then look at the necessity for adding a `Dropdown.Input` etc. component

More detail/thoughts below, but not essential to read.

> Currently the typehead also blocks outter inputs from being used while the drop-down options are open. Just wanted to share this in case it might change an approach you might take with this.

@Aaron-Conroy @rluders I found this as well with tests I did with the component. This is due to a listener in the typeahead. So even if the Dropdown were to have header or footer sections that are outside of the `FloatingFocusManager` or `FloatingList`, it will not fix the issue with inputs. Additionally, if we move anything outside of the `FloatingList` it will no longer have access to `UseListNavigation` (arrow navigation) and would suddenly have to be tabbed to. I also ran into one other edge case with `UseListNavigation` that doesn't seem intuitive.

I'm wondering if only allowing arrow/typeahead navigation (which sacrifice native tab navigation behaviour in the dropdown) will always be the best UX for all use cases of this component. Maybe these should be opt in? After looking at this more in depth, I'm back to feeling like `UseTypeahead` cons may outweigh the pros. To me its a 'nice to have', but can cause as much UX/dev as it does benefit.

nigellima:
> TL;DR After looking at this and Floating UI more in depth, my feeling now is:
> 
> * `UseTypeahead` should be removed or opt-in via props on the Dropdown
> * if `UseTypeahead` is kept opt-in, it will need a way to be toggled off if there is to be any sort of input is allowed in a dropdown (adding `Header` or `Footer` will not fix the issues, and might create more navigation issues with `UseListNavigation`).
> * an idea for toggling off `UseTypeahead` would be to check if an input is currently focussed (something like document.activeElement?.tag === 'INPUT') but I would be interested if there were a suggestion for a better way to do a toggle
> * `UseListNavigation` should possibly also be opt-in as it always overrides (does not allow) native tab navigation; depending on the rest of the UI, this may not always be consistent/intuitive
> * if the `Dropdown` is allowing focus to leave (as it currently does with `modal={false}` prop, there will be the issue that @Aaron-Conroy ran into with `UseTypeahead` blocking inputs elsewhere in the UI
> 
> So, my proposal would be (obviously open to suggestions):
> 
> * [ ]  remove `UseTypeahead`
> * [ ]  make `UseListNavigation` opt-in via props
> * [ ]  consider changing how `Dropdown.Item` is composed (maybe passing props rather than allowing children to put some guardrails on ways the component cannot be used)
> * [ ]  then look at the necessity for adding a `Dropdown.Input` etc. component
> 
> More detail/thoughts below, but not essential to read.
> 
> > Currently the typehead also blocks outter inputs from being used while the drop-down options are open. Just wanted to share this in case it might change an approach you might take with this.
> 
> @Aaron-Conroy @rluders I found this as well with tests I did with the component. This is due to a listener in the typeahead. So even if the Dropdown were to have header or footer sections that are outside of the `FloatingFocusManager` or `FloatingList`, it will not fix the issue with inputs. Additionally, if we move anything outside of the `FloatingList` it will no longer have access to `UseListNavigation` (arrow navigation) and would suddenly have to be tabbed to. I also ran into one other edge case with `UseListNavigation` that doesn't seem intuitive.
> 
> I'm wondering if only allowing arrow/typeahead navigation (which sacrifice native tab navigation behaviour in the dropdown) will always be the best UX for all use cases of this component. Maybe these should be opt in? After looking at this more in depth, I'm back to feeling like `UseTypeahead` cons may outweigh the pros. To me its a 'nice to have', but can cause as much UX/dev as it does benefit.

Hi Chris. Thanks for digging into that.

So I believe in that case I would rather narrow down the support for what can be done within the Dropdown rather than start adding toggles to its a11y features.

All these features should be spread to `Popover`, `Combobox`, and `Dropdown` components accordingly, just like we see in shadncn, headlessui, etc. All of these other libraries have the navigation and typehead builtin (and I believe there is no way to turn it off).

What do you think?

ChrisCoastal:
@nigellima My pleasure, thanks for the reply. That makes sense to me, as long as the Dropdown is fairly prescriptive on how it's used. I think we may want to set modal={true} so in most cases the dropdown won't cause unexpected behaviour if focus leaves. And maybe also only allow text/icons as Dropdown.Items (radio buttons may work, though I'm doubtful, I'll need to double check). What do you think?

nigellima:
> 



> @nigellima My pleasure, thanks for the reply. That makes sense to me, as long as the Dropdown is fairly prescriptive on how it's used. I think as long as we set closeOnFocusOut on the focusmanager to I mount the typeahead if the user navigates away. And maybe also only allow text/icons as Dropdown.Items (radio buttons may work, though I'm doubtful, I'll need to double check). What do you think?

Nice. Yeah, to me it looks like the right path... then there is room here for the implementation of `Popover` and `Combobox` which would support these additional use cases with inputs.

Regarding the Dropdown.Items, I believe we could separate what is `Dropdown.Item`, Dropdown.ItemCheckbox`, and `Dropdown.ItemRadio`. I have seen this kinda of API in other places too.

Do you guys have any further input? @rluders @tulup-conner

rluders:
> > 
> 
> > @nigellima My pleasure, thanks for the reply. That makes sense to me, as long as the Dropdown is fairly prescriptive on how it's used. I think as long as we set closeOnFocusOut on the focusmanager to I mount the typeahead if the user navigates away. And maybe also only allow text/icons as Dropdown.Items (radio buttons may work, though I'm doubtful, I'll need to double check). What do you think?
> 
> Nice. Yeah, to me it looks like the right path... then there is room here for the implementation of `Popover` and `Combobox` which would support these additional use cases with inputs.
> 
> Regarding the Dropdown.Items, I believe we could separate what is `Dropdown.Item`, `Dropdown.ItemCheckbox`, and `Dropdown.ItemRadio`. I have seen this kinda of API in other places too.
> 
> Do you guys have any further input? @rluders @tulup-conner

Yeah, I agree. I think that this seems to be reasonable. BTW, nice dive into the component structure and behavior @ChrisCoastal.

Aaron-Conroy:
Do we have any updates on this? Been quite a while since we've had an update

ChrisCoastal:
Hey @Aaron-Conroy @rluders. Apologies, I had to set this down while getting caught up on other work. Adding in the checkbox and radio items added some complexity, but I've gotten them both working.

https://github.com/themesberg/flowbite-react/assets/91000890/a93195a9-41f5-4d5b-98fd-f6c899ce1ad6

I'll hopefully have time to update the pull request later; still requires some cleanup. Also wondering how/if the selected input state should be exposed by the component? Is this something that you want the component to handle?

Also note that this change will not fix any issue with consumers adding text inputs (will leave this to the documentation as previously discussed). Appreciate the patience.

ChrisCoastal:
@rluders @nigellima I've opened a new pull request with the proposed changes #1095 to better follow the style guidelines and give more clarity to the scope. Closing this request and will continue with any further work over there.

PR #949, opened 2023-09-08
feat: add `<Banner>` component

tulup-conner:

Adds the [Banner](https://flowbite.com/docs/components/banner/) component from the main Flowbite library.

This component is currently a work-in-progress; you can expect that there may be significant changes in the future, in particular, to its theme. There currently isn't a theme, but there probably will be in the future.

Unit tests and Storybook stories will be added in the future.

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -5,7 +5,7 @@ export type BannerComponentProps = ComponentProps<'div'>;
 
 const BannerComponent: FC<BannerComponentProps> = ({ children, ...props }) => {
   return (
-    <div data-testid="flowbite-banner" tabIndex={-1} {...props}>
+    <div data-testid="flowbite-banner" role="banner" tabIndex={-1} {...props}>
       {children}
     </div>
   );
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -1,0 +1,17 @@
+import type { ComponentProps, FC } from 'react';
+import { BannerCollapseButton } from './BannerCollapseButton';
+
+export type BannerComponentProps = ComponentProps<'div'>;
+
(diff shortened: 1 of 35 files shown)


src/components/Banner/Banner.tsx
```diff
@@ -0,0 +1,17 @@
+import type { ComponentProps, FC } from 'react';
+import { BannerCollapseButton } from './BannerCollapseButton';
+
+export type BannerComponentProps = ComponentProps<'div'>;
+
+const BannerComponent: FC<BannerComponentProps> = ({ children, ...props }) => {
+  return (
+    <div data-testid="flowbite-banner" tabIndex={-1} {...props}>
```
nigellima, line 8: Add a `role="banner"` here for screen readers

tulup-conner: Thank you, doing this now. Also, this isn't in the vanilla library @zoltanszogyenyi

src/components/Banner/BannerCollapseButton.tsx
```diff
@@ -0,0 +1,22 @@
+import type { FC, MouseEventHandler } from 'react';
+import type { ButtonProps } from '../Button';
+import { Button } from '../Button';
+
+export type BannerCollapseButtonProps = ButtonProps;
+
+export const BannerCollapseButton: FC<BannerCollapseButtonProps> = ({ children, ...props }) => {
+  const onClick: MouseEventHandler<HTMLButtonElement> = (event) => {
+    const collapseButton = event.target as HTMLButtonElement;
+    const parentBanner = collapseButton.closest('[data-testid="flowbite-banner"]');
```
nigellima, line 10: I'm not sure if using the data-testid here for implementation logic is advisable.

Maybe use the `role=banner` here to identify the parent

tulup-conner: Agree now that we have that, will change that as well now.

PR #987, opened 2023-09-23
fix(toggleswitch): disable + checked state and deprecated html event replacement

### LINKED ISSUE #986: ToggleSwitch checked={true} disabled={true} not visible

## Steps to reproduce

1. Create a toggle switch and set checked={true} and disabled={true}

## Current behavior

- cheked and disabled
image

- not checked and disabled:
image

## Expected behavior

It should be visible showing its status as checked.

rluders:

As described in issue #986 the component `ToggleSwitch` wasn't acting as expected when `checked={true} disabled={true}`. This PR fixes the issue. This also replaces the deprecated `onKeyPress` API with `onKeyDown`.

Fixes #986 

⚠️ This PR replaces the `onKeyPress` with `onKeyDown`. It should not break anything, since the API is only internally used.

### DIFF
--- a/src/components/ToggleSwitch/ToggleSwitch.tsx
+++ b/src/components/ToggleSwitch/ToggleSwitch.tsx
@@ -1,5 +1,5 @@
 import type { ComponentProps, FC, KeyboardEvent, MouseEvent } from 'react';
-import { useId } from 'react';
+import { useId, useRef } from 'react';
 import { twMerge } from 'tailwind-merge';
 import type { DeepPartial, FlowbiteBoolean, FlowbiteColors } from '../../';
 import { useTheme } from '../../';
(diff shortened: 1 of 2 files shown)


SutuSebastian (commented):
Should we take into account throttling? Or this does not cause any issues in this case, if u simply press and hold a button.

src/components/ToggleSwitch/ToggleSwitch.tsx
```diff
@@ -44,16 +44,18 @@ export const ToggleSwitch: FC<ToggleSwitchProps> = ({
 }) => {
   const id = useId();
   const theme = mergeDeep(useTheme().theme.toggleSwitch, customTheme);
+  const buttonRef = useRef<HTMLButtonElement>(null);
```
SutuSebastian, line 47: isn't this redundant not being used anywhere? 🤔

rluders: Yes. Dead code. Thanks for noticing.

SutuSebastian (approved):

PR #1034, opened 2023-10-01
feat: indicator Component

MateoWartelle:

Indicator component to match 
https://flowbite.com/docs/components/indicators/


rluders:
This PR will also be required to update some other components. I can only remember the `Avatar` right now, but there are probably some others.

MateoWartelle:
> This PR will also be required to update some other components. I can only remember the `Avatar` right now, but there are probably some others.

Yes I couldn't get it to work on this example : 

https://flowbite-react-l570u8y3f-themesberg.vercel.app/docs/components/indicators#status-indicator

MateoWartelle:
> This PR will also be required to update some other components. I can only remember the `Avatar` right now, but there are probably some others.

What specifically?

rluders:
> > This PR will also be required to update some other components. I can only remember the `Avatar` right now, but there are probably some others.
> 
> What specifically?

https://github.com/themesberg/flowbite-react/blob/1189d3d16981ba4405f450ddfd242df4d91576b9/src/components/Avatar/Avatar.tsx#L147

The `Avatar` component has an inner indicator that is used only for it. With the `Indicator` component, it would be nice to use it for the component or any other approach that I'm not considering here.

PR #1036, opened 2023-10-01
Fix/lint warning on button test file

paghar:

### DIFF
--- a/src/components/Button/Button.spec.tsx
+++ b/src/components/Button/Button.spec.tsx
@@ -171,7 +171,7 @@ describe('Components / Button', () => {
       it('should render as button `as={null}`', () => {
         render(
           <ul>
-            <Button as={null as null} label="Something or other" />
+            <Button as={null} label="Something or other" />
           </ul>,
         );
 
--- a/src/components/Button/Button.spec.tsx
+++ b/src/components/Button/Button.spec.tsx
@@ -171,7 +171,7 @@ describe('Components / Button', () => {
       it('should render as button `as={null}`', () => {
         render(
           <ul>
-            <Button as={null as any} label="Something or other" />
+            <Button as={null as null} label="Something or other" />
           </ul>,
         );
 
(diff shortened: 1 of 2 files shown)


paghar:
@rluders I have changed the type to fix lint warning.
please check that🙏

rluders:
@nigellima WDYT?

src/components/Button/Button.spec.tsx
```diff
@@ -171,7 +171,7 @@ describe('Components / Button', () => {
       it('should render as button `as={null}`', () => {
         render(
           <ul>
-            <Button as={null as any} label="Something or other" />
+            <Button as={null as null} label="Something or other" />
```
nigellima, line 174: @paghar is the type casting here necessary as now you upgraded the `as` prop to `as?: T | null`?

hamed-dehghan: @nigellima Should I remove it from the examples?

paghar: @nigellima 
should I remove example from the documentation?

paghar:
@rluders should I close this PR?

rluders:
> @rluders should I close this PR?

I actually missed this one, didn't have time to test it. Maybe @SutuSebastian could also give some feedback here.

src/components/Button/Button.spec.tsx
SutuSebastian, line 174: `as={null}` is sufficient here

SutuSebastian:
My only question is:
- is this nullable approach just to be able to test that it should render the default html element (in this case `button`)?

src/components/Button/Button.spec.tsx
paghar: You are right

paghar:
> My only question is:
> 
>     * is this nullable approach just to be able to test that it should render the default html element (in this case `button`)?

Yes, it seems to be for this reason.

SutuSebastian:
Then we're all good here

paghar:
@rluders

Issue #1020, opened 2023-09-29
Button with a label


PadowYT2:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Steps to reproduce

1. Make a button
2. Assign a label

## Current behavior

The label doesn't exist on the button

## Expected behavior

The label should be added to the button

## Context

https://www.flowbite-react.com/docs/components/button#button-with-label

rluders:
The issue is probably here:
https://github.com/themesberg/flowbite-react/blob/c3bf704dd9775390a5efac638d84192b588fcbbd/src/components/Button/Button.tsx#L153

Please, introduce a unit test to cover the scenario to avoid future regressions.

MateoWartelle:
> The issue is probably here:
> 
> https://github.com/themesberg/flowbite-react/blob/c3bf704dd9775390a5efac638d84192b588fcbbd/src/components/Button/Button.tsx#L153
> 
> Please, introduce a unit test to cover the scenario to avoid future regressions.

If both children and label are set what should happen?

`<Button label="2">
      Messages
    </Button>`

<img width="154" src="https://github.com/themesberg/flowbite-react/assets/27743176/43510dd1-ad27-4165-9f48-cf369af70838">

rluders:
https://flowbite.com/docs/components/buttons/#button-with-label

PR #1041, opened 2023-10-03
fix: button with a label

MateoWartelle:

#1020 

change check to show children and label independently

### DIFF
--- a/src/components/Button/Button.tsx
+++ b/src/components/Button/Button.tsx
@@ -149,9 +149,8 @@ const ButtonComponentFn = <T extends ElementType = 'button'>(
               {processingSpinner || <Spinner size={size} />}
             </span>
           )}
-          {typeof children !== 'undefined' ? (
-            children
-          ) : (
+          {children}
+          {label && (
             <span data-testid="flowbite-button-label" className={twMerge(theme.label)}>
               {isProcessing ? processingLabel : label}
             </span>
--- a/src/components/Button/Button.spec.tsx
+++ b/src/components/Button/Button.spec.tsx
@@ -118,6 +118,24 @@ describe('Components / Button', () => {
       expect(button()).toHaveTextContent('Something or other');
     });
 
+    it('should render children then label', () => {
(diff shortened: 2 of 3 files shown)


src/components/Button/Button.tsx
```diff
@@ -148,12 +148,13 @@ const ButtonComponentFn = <T extends ElementType = 'button'>(
...
-          {typeof children !== 'undefined' ? (
-            children
-          ) : (
+          {typeof children !== 'undefined' ? children : ''}
+          {typeof label !== 'undefined' ? (
             <span data-testid="flowbite-button-label" className={twMerge(theme.label)}>
               {isProcessing ? processingLabel : label}
             </span>
+          ) : (
+            ''
```
rluders, line 157: `null`

src/components/Button/Button.tsx
```diff
@@ -148,12 +148,13 @@ const ButtonComponentFn = <T extends ElementType = 'button'>(
               {processingSpinner || <Spinner size={size} />}
             </span>
           )}
-          {typeof children !== 'undefined' ? (
-            children
-          ) : (
+          {typeof children !== 'undefined' ? children : ''}
+          {typeof label !== 'undefined' ? (
```
rluders, line 152: What would happen if `label=""` or `<Button label>Test</Button>`?

rluders (changes requested):

src/components/Button/Button.tsx
```diff
@@ -148,9 +148,8 @@ const ButtonComponentFn = <T extends ElementType = 'button'>(
               {processingSpinner || <Spinner size={size} />}
             </span>
           )}
-          {typeof children !== 'undefined' ? (
-            children
-          ) : (
+          {children ?? children}
```
SutuSebastian, line 151: if `children` is simply rendering itself, then this is redundant.

Instead, simply render it like `{children}`, react will take care if its `null` or `undefined` and not render anything, else render the node.

PR #1053, opened 2023-10-09
Fix/fetch response error handling

paghar:

### DIFF
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -1,0 +1,6 @@
+export const safeResJson = (res: Response) => {
+  if (res.ok) {
+    return res.json();
+  }
+  throw new Error('Internal server error!');
+};
--- a/app/page.tsx
+++ b/app/page.tsx
@@ -9,6 +9,7 @@ import { HiClipboardCopy, HiOutlineArrowRight } from 'react-icons/hi';
 import '~/app/docs.css';
 import '~/app/style.css';
 import { Button, Flowbite, Footer, Navbar, TextInput, Tooltip } from '~/src';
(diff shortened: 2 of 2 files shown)


paghar:
@rluders Hi
Thank you for checking the changes, I will always be happy to help.

src/helpers/http.ts
```diff
@@ -0,0 +1,7 @@
+export const safeResJson = (res: Response) => {
+  if (res.ok) {
+    return res.json()
+  } else {
```
rluders, line 4: You don't need the `else` statement. Just throw the exception after the `if` block.

rluders (changes requested):
I'm unsure if I understood what it is trying to solve, but... I guess it doesn't hurt to get it merged after the change.

src/helpers/http.ts
paghar: you are right.
fixed now👌

rluders:
Looks good @paghar. But the prettier doesn't seem happy.

paghar:
@rluders Sorry about that.
I fixed the problem, but I'm unable to update branch with rebase because of a conflict!
So I merged It....Is that OK?

rluders:
> @rluders Sorry about that. I fixed the problem, but I'm unable to update branch with rebase because of a conflict! So I merged It....Is that OK?

You need to rebase... If you are having issues with the rebase, I can try to help tomorrow, just let me know if you didn't get into solving the conflict when rebasing.

paghar:
@rluders It is fixed now.
please check that.

rluders (approved):

PR #1056, opened 2023-10-11
feat: `ThemeModeScript` - avoid page flicker on reload on SSR application (NextJS, Remix)

SutuSebastian:

There is a flicker between `light->dark` theme on full page reload, thats due to SSR client hydration of the react components, and since the document `html` is being updated only after NextJS mounts, flicker occurs leading to bad UX.

### Changes

- [x] create `ThemeModeScript` to avoid page flicker on reload
- [x] add script to root layout on homepage + docs

### Usage

NextJS App router

```tsx
// app/layout.tsx

export default function RootLayout ({ children }) {
  return (
    <html>
      <head>
        <ThemeModeScript />
      </head>
      <body>
        {children}
      </body>
    </html>
  );
};
```

NextJS Pages router

```tsx
// pages/_document.tsx

export default function Document () {
  return (
    <Html>
      <Head>
        <ThemeModeScript />
      </Head>
      <body>
        <Main />
        <NextScript />
      </body>
    </Html>
  );
}
```

### Before

https://github.com/themesberg/flowbite-react/assets/41998826/3584df2f-4722-48aa-8acc-6a2f67d43ba9

### After

https://github.com/themesberg/flowbite-react/assets/41998826/b62a2ea3-a559-4a72-bc4c-e2155fef4ca7


src/theme-mode-script.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
```
rluders, line 1: I guess that I'll only hope to see an update on the documentation as well.

https://www.flowbite-react.com/docs/customize/dark-mode

rluders (changes requested):

src/theme-mode-script.tsx
SutuSebastian: Good point, let me see what I can do.

src/theme-mode-script.tsx
```diff
@@ -0,0 +1,42 @@
...
+      const mode = window.localStorage.getItem('${localStorageKey}') ?? '${mode}' ?? '${defaultMode}';
+      const computedMode =
+        mode === 'auto' ? (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'dark' : 'light') : mode;
+
+      if (computedMode === 'dark') {
+        document.documentElement.classList.add('dark');
+      } else {
+        document.documentElement.classList.remove('dark');
+      }
+    } catch (e) {}
```
nigellima, line 40: Maybe we should log this? It could be helpful for users to report the error if anything happens

SutuSebastian: The try catch block is solely here for the SSR side, in case it tries to execute this on server, it doesn't work.

src/theme-mode-script.tsx
SutuSebastian: Docs will come out right before we release, so there is no blocker if any other fix touches docs.

rluders: ack

PR #1079, opened 2023-10-20
Fix/remove unneccesary dependency array

paghar:

### DIFF
--- a/src/helpers/use-watch-localstorage-value.ts
+++ b/src/helpers/use-watch-localstorage-value.ts
@@ -19,5 +19,5 @@ export const useWatchLocalStorageValue = ({
   useEffect(() => {
     window.addEventListener('storage', handleStorageChange);
     return () => window.removeEventListener('storage', handleStorageChange);
-  });
+  }, []); // eslint-disable-line
 };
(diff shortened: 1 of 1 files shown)


paghar:
@rluders Please take a look at this, thanks....

SutuSebastian:
@paghar I wouldn't really call it "unnecessary" tho, It is there more for performance because we want to mount the listeners once, not every single render. One single downside that I can think of is the callback closure and possibly stale/outdated callback variables if they have from one execution to another, but here with this implementation, it is not the case.

rluders:
Agreed with @SutuSebastian. I think that we don't need this. But, thanks anyway. :)

paghar:
@rluders @SutuSebastian Thanks for the explanation, Is there any better solution to get rid of that warning?

rluders:
@paghar in this case, I think that we can just ignore it.

rluders:
@paghar I'll reopen the PR so, maybe you can just add the line to ignore the warning and it should be all.

paghar:
> @paghar I'll reopen the PR so, maybe you can just add the line to ignore the warning and it should be all.

Thanks @rluders 🙏

paghar:
Done @rluders,
Thank you...

SutuSebastian:
> @rluders @SutuSebastian Thanks for the explanation, Is there any better solution to get rid of that warning?

U're most welcome! Sure, there is a solution, that is putting the callback `handleStorageChange` in the dependency array, but this means that I **have** to memoise the callback everywhere I use it, its a bummer 🙄.

PR #1080, opened 2023-10-20
Fix/disable eslint warning on floating label

paghar:

### DIFF
--- a/src/components/FloatingLabel/FloatingLabel.tsx
+++ b/src/components/FloatingLabel/FloatingLabel.tsx
@@ -4,12 +4,30 @@ import { twMerge } from 'tailwind-merge';
 import { mergeDeep } from '../../helpers/merge-deep';
 import { getTheme } from '../../theme-store';
 import type { DeepPartial } from '../../types';
-import type { FlowbiteSizes } from '../Flowbite';
+import type { FlowbiteColors, FlowbiteSizes } from '../Flowbite';
+
+export interface FlowbiteFloatingLabelHelperText extends Partial<FlowbiteColors> {
+  default: string
+  success: string
+  error: string
+}
 
 export interface FlowbiteFloatingLabelTheme {
-  input: any; // eslint-disable-line
-  label: any; // eslint-disable-line
-  helperText: any; // eslint-disable-line
+  input: {
+    [key: string]: {
+      [key: string]: {
+        [key: string]: string
+      }
+    }
+  };
+  label: {
+    [key: string]: {
+      [key: string]: {
+        [key: string]: string
+      }
+    }
+  };
+  helperText: FlowbiteFloatingLabelHelperText;
 }
 
 export interface FlowbiteFloatingLabelSizes extends Pick<FlowbiteSizes, 'sm' | 'md' | 'lg'> {
--- a/src/components/FloatingLabel/FloatingLabel.tsx
+++ b/src/components/FloatingLabel/FloatingLabel.tsx
@@ -7,9 +7,9 @@ import type { DeepPartial } from '../../types';
 import type { FlowbiteSizes } from '../Flowbite';
 
 export interface FlowbiteFloatingLabelTheme {
-  input: any;
-  label: any;
-  helperText: any;
+  input: any; // eslint-disable-line
+  label: any; // eslint-disable-line
+  helperText: any; // eslint-disable-line
 }
 
 export interface FlowbiteFloatingLabelSizes extends Pick<FlowbiteSizes, 'sm' | 'md' | 'lg'> {


paghar:
@rluders I tried to make annoying warnings disappear by detecting and defining the actual types, but it seems to be tricky...
so I disabled the eslint warning here...

src/components/FloatingLabel/FloatingLabel.tsx
```diff
@@ -7,9 +7,9 @@ import type { DeepPartial } from '../../types';
 import type { FlowbiteSizes } from '../Flowbite';
 
 export interface FlowbiteFloatingLabelTheme {
-  input: any;
-  label: any;
-  helperText: any;
+  input: any; // eslint-disable-line
+  label: any; // eslint-disable-line
+  helperText: any; // eslint-disable-line
```
rluders, line 12: This is probably: FlowbiteColors

src/components/FloatingLabel/FloatingLabel.tsx
```diff
@@ -7,9 +7,9 @@ import type { DeepPartial } from '../../types';
 import type { FlowbiteSizes } from '../Flowbite';
 
 export interface FlowbiteFloatingLabelTheme {
-  input: any;
-  label: any;
-  helperText: any;
+  input: any; // eslint-disable-line
```
rluders, line 10: I think that for both: `input` and `label` it would be necessary to create a type.

rluders (changes requested):

src/components/FloatingLabel/FloatingLabel.tsx
paghar: witch one do you mean @rluders?

rluders: Sorry, the `helperText` is the `FlowbiteColors` type. :)

paghar: @rluders 
the type `FlowbiteColors ` is like this:

```
export interface FlowbiteColors extends FlowbiteStateColors {
  [key: string]: string;
  blue: string;
  cyan: string;
  dark: string;
  gray: string;
  green: string;
  indigo: string;
  light: string;
  lime: string;
  pink: string;
  purple: string;
  red: string;
  teal: string;
  yellow: string;
}
```

while the `helperText` is used like this:
```
  helperText: {
    default: 'mt-2 text-xs text-gray-600 dark:text-gray-400',
    success: 'mt-2 text-xs text-green-600 dark:text-green-400',
    error: 'mt-2 text-xs text-red-600 dark:text-red-400',
  },
```

this will cause error because colors are not optional in `FlowbiteColors`

paghar: maybe

rluders: I think that it would be fine to use `FlowbiteColors`. It extends the `FlowbiteStateColors` that has all the states, except by the `default`, and I don't think that it would be a problem.

paghar: @rluders 
image
Are you sure this is OK?

rluders: Did you change the https://github.com/themesberg/flowbite-react/blob/0f7ca75d5709eded9b9a1c9e406fc28962aba7ad/src/components/FloatingLabel/FloatingLabel.tsx#L33C1-L33C18 to be:

```
helperText: keyof FlowbiteColors;
```

??

rluders: (BTW: Not saying that this PR is going to solve it, but we need to put all the colors into some standard interface - the colors patterns are becoming really complicated to manage)

paghar: > keyof

after doing that, this is the result:
image

rluders: @paghar if necessary you can create a specific type for the `FlowbiteFloatingLabelHelperText` (oh my... look at this name) extending the `FlowbiteColors` and picking only the ones that you want. There are some components (Button?) that do something similar.

paghar:
@rluders I extended `FlowbiteColors` as mentioned and used this type for both `input` and `label` keys:
```
{
    [key: string]: {
      [key: string]: {
        [key: string]: string
      }
    }
}
```

paghar:
@rluders Sorry to bother you. Is there anything here that I can help?

src/components/FloatingLabel/FloatingLabel.tsx
```diff
@@ -4,12 +4,30 @@ import { twMerge } from 'tailwind-merge';
...
+  success: string
+  error: string
+}
 
 export interface FlowbiteFloatingLabelTheme {
-  input: any;
-  label: any;
-  helperText: any;
+  input: {
+    [key: string]: {
```
rluders, line 17: I think that you could have created a type for this one since it is used in the +1 place

rluders (changes requested):

rluders:
@SutuSebastian @Ncookiez @nigellima could you folks take a look into this PR?

SutuSebastian:
> @SutuSebastian @Ncookiez @nigellima could you folks take a look into this PR?

needs more work when it comes to typing those theme interfaces

SutuSebastian:
deprecated by #1118

PR #1090, opened 2023-10-26
Update Banner.tsx

nikcode9:
enhanced.

### DIFF
--- a/src/components/Banner/Banner.tsx
+++ b/src/components/Banner/Banner.tsx
@@ -1,9 +1,11 @@
-import type { ComponentProps, FC } from 'react';
-import { BannerCollapseButton } from './BannerCollapseButton';
+import React, { FC, ReactNode } from 'react';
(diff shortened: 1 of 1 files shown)


rluders:
@nikcode9 can you give more details and explain the goals here?

rluders:
@nikcode9 and what are the benefits here? 🤔 @nigellima @SutuSebastian

SutuSebastian:
Using typescript and prop-types at the same time is redundand.

No typescript -> prop types
Typescript -> no prop types

Other than that, I dont see much value being added here, this is not even a library component, it is just used for the docs, no need for fancy JSDoc or display name IMO

SutuSebastian:
Also, we lost all native "div" element prop types

rluders:
Okay. So, as I was thinking this is a nope. Thanks anyway, @nikcode9.

PR #1095, opened 2023-10-30
feat(dropdownitems): add DropdownItemCheckbox and DropdownItemRadio

ChrisCoastal:
re #916 and #941 

This commit follows from the conversation in #941 and #916 - adding functionality to the Dropdown component to include Dropdown.ItemRadio and Dropdown.ItemCheckbox.
It also fixes a bug in DropdownItem.tsx that will only ever set the [selectIndex to `null`](https://github.com/themesberg/flowbite-react/blob/8ddebe16f57500d0e55e71e1b5fe42d263fa9767/src/components/Dropdown/DropdownItem.tsx#L49C31-L49C49). Currently, however, [selectedIndex](https://github.com/themesberg/flowbite-react/blob/8ddebe16f57500d0e55e71e1b5fe42d263fa9767/src/components/Dropdown/Dropdown.tsx#L127C10-L127C23) isn't used anywhere, so possibly it should just be removed (or maybe there is conditional styling that should be added to reflect its state)?

Related to #916 and #941

There are no breaking API changes. Both FloatingUI Typeahead functionality and Arrow Navigation will work for the new item components (tabbing through the dropdown also works) in accordance with WCAG.

Additional considerations:
- Storybook will still need to be updated to reflect the new component options (I wasn't able to get the Storybook dev server running, but am happy to do that work as well if the changes look good).
- These changes don't address any issues Flowbite users are running into when trying to place text inputs as children of `DropdownItem` (and the FloatingUI Typeahead catching all the input keystrokes). The component could possibly be updated to not accept children or offer a warning in development.

### DIFF
--- a/content/docs/components/dropdown.mdx
+++ b/content/docs/components/dropdown.mdx
@@ -37,6 +37,23 @@ Add custom icons next to the menu items by using the `icon` prop on the `<Dropdo
 
 <Example name="dropdown.itemsWithIcon" />
 
+## Dropdown items with radio input
+
+Add radio items to the dropdown with `<Dropdown.ItemRadio>` component.
+`<Dropdown.ItemRadio>` accepts a ref that will be passed to the underlying `Radio` component's input element.
+`<Dropdown.ItemRadio>` components with the same `name` prop will be treated as a radio group.
+Optionally pass `dismissOnClick={false}` to the parent `Dropdown` to keep the dropdown open after a radio is toggled.
+
+<Example name="dropdown.itemRadio" />
+
+## Dropdown items with checkbox input
+
+Add checkbox items to the dropdown with `<Dropdown.ItemCheckbox>` component.
+`<Dropdown.ItemCheckbox>` accepts a ref that will be passed to the underlying `Checkbox` component's input element.
+Optionally pass `dismissOnClick={false}` to the parent `Dropdown` to keep the dropdown open after a checkbox is checked.
+
+<Example name="dropdown.itemCheckbox" />
+
 ## Inline dropdown
 
 Use the `inline` prop to make the dropdown appear inline with the trigger element.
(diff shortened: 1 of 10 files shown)


rluders:
@ChrisCoastal looks promising. Could you please rebase it with master and also add an example to the documentation? Thanks! :)

rluders:
@ChrisCoastal gently ping. :)

ChrisCoastal:
Thanks for the gentle bump @rluders. I've done the rebase and updated the docs. Please let me know what you think and if I can make any further changes. 🙃🙏

tulup-conner:
I'm not a fan of this idea, personally. I think the best direction we could take here is to revert `<Dropdown.Item>` to `<li>` and allow the user to create whatever they want. There *is* a spec on flowbite.com for Dropdowns with checkboxes and Dropdowns with radios, but I think we can easily achieve that by adding a `<Checkbox/>` inside each `<Dropdown.Item>`. We can provide the simple JSX template for that as opposed to implementing functionality to fully control them.

ChrisCoastal:
@tulup-conner I agree with you. I think that it is maybe best not to implement the Dropdown.Checkbox and Dropdown.Radio in the React lib at all. As long as Floating UI is being used (for Typeahead and Focus Trap) element within each `<li>` that Floating interacts with must be a `<button>` and `<input>` should not be placed within `<button>` according to the W3C spec. 

Honestly, as long as Floating UI is being used as it is now, my feeling would be that the `Dropdown.Item` component _should not accept children at all_, but only props that define text and/or icon within. This would put guardrails on how the component is consumed. I think this is important as this is actually at the heart of the original issue that started this - devs trying to put `<input type='text'>` within the Dropdown.Item component and wondering why the text field is unresponsive. 

Whether or not that kind of implementation is consistent with the sort of API you want for the lib (passing props rather than children), I'm not sure.

PR #1109, opened 2023-11-07
refactor(components): separate `AdvancedRating` from `Rating` theming

SutuSebastian:

### Changes

- [x] separate theme  `RatingAdvanced` from `Rating` since they cant ever be coupled

### API changes

- removed `advanced` from `FlowbiteRatingTheme` interface

### Result

Before
Screenshot 2023-11-07 at 10 03 10

After
Screenshot 2023-11-07 at 09 57 50

### DIFF
--- a/src/components/Rating/RatingAdvanced.tsx
+++ b/src/components/Rating/RatingAdvanced.tsx
@@ -26,7 +26,7 @@ export const RatingAdvanced: FC<RatingAdvancedProps> = ({
   theme: customTheme = {},
   ...props
 }) => {
-  const theme = mergeDeep(getTheme().rating.advanced, customTheme);
+  const theme = mergeDeep(getTheme().ratingAdvanced, customTheme);
 
   return (
     <div className={twMerge(theme.base, className)} {...props}>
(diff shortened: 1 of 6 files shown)


src/components/Rating/RatingAdvanced.tsx
```diff
@@ -26,7 +26,7 @@ export const RatingAdvanced: FC<RatingAdvancedProps> = ({
   theme: customTheme = {},
   ...props
 }) => {
-  const theme = mergeDeep(getTheme().rating.advanced, customTheme);
+  const theme = mergeDeep(getTheme().ratingAdvanced, customTheme);
 
   return (
     <div className={twMerge(theme.base, className)} {...props}>
```
rluders, line 32: I'm wondering why don't call it `root.base` here. And I'm not saying that it is wrong... But actually, just thinking why we sometimes have it as root, and sometimes we only use `.base`. 🤔

SutuSebastian: So I noticed (and it makes sense) `root` is used when there is component composition (eg: `Rating.Star`) and `RatingStar` component lives under a separate umbrella, but still u want to control the theme granularity from the **root** component, therefore whenever a component hosts many other under its umbrella, we have a `root`.

Hopefully I put that in a more clear perspective 🤔

rluders (approved):

PR #1113, opened 2023-11-07
Fix/define type for json responses

paghar:

### DIFF
--- a/src/helpers/http.ts
+++ b/src/helpers/http.ts
@@ -1,4 +1,8 @@
-export const safeResJson = (res: Response) => {
+export interface JsonResponseType {
+  [key: string]: string
+}
+
+export const safeResJson = (res: Response): Promise<JsonResponseType> => {
   if (res.ok) {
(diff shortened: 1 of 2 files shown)


paghar:
Hi👋🏻@rluders 
This PR is merge ready

src/helpers/http.ts
```diff
@@ -1,4 +1,8 @@
-export const safeResJson = (res: Response) => {
+export interface JsonResponseType {
```
SutuSebastian, line 1: thing is, we don't always know if the response is a plain `Record`, basically it can be anything.
I left it `any` because I needed to strongly type the responses of all API calls and I just didn't feel like doing it, derp.
Normally it should be typed as `Promise<unknown>` and `unknown` also extends `T`, which comes from the place where the function is fired.

@rluders what do u think about this? should we just merge the result to be "fale always" a `Record`?

SutuSebastian:
deprecated by #1118

PR #1121, opened 2023-11-09
Added draggable option to carousel component

Sigabiel:

Edited the Carousel.tsx to add the option to disable the mouse drag of the carousel component

Discussion (https://github.com/themesberg/flowbite-react/discussions/1120)

Optional Prop "draggable":boolean for carousel component.

### DIFF
--- a/src/components/Carousel/Carousel.tsx
+++ b/src/components/Carousel/Carousel.tsx
@@ -50,6 +50,7 @@ export interface CarouselProps extends ComponentProps<'div'> {
   indicators?: boolean;
   leftControl?: ReactNode;
   rightControl?: ReactNode;
+  draggable?: boolean;
   slide?: boolean;
   slideInterval?: number;
   theme?: DeepPartial<FlowbiteCarouselTheme>;
@@ -67,6 +68,7 @@ export const Carousel: FC<CarouselProps> = ({
   leftControl,
   rightControl,
   slide = true,
+  draggable = true,
   slideInterval,
   className,
   theme: customTheme = {},
@@ -145,10 +147,10 @@ export const Carousel: FC<CarouselProps> = ({
     >
       <ScrollContainer
         className={twMerge(theme.scrollContainer.base, (isDeviceMobile || !isDragging) && theme.scrollContainer.snap)}
-        draggingClassName="cursor-grab"
+        draggingClassName={draggable ? "cursor-grab" : ""}
         innerRef={carouselContainer}
         onEndScroll={handleDragging(false)}
-        onStartScroll={handleDragging(true)}
+        onStartScroll={handleDragging(draggable)}
         vertical={false}
       >
         {items?.map((item, index) => (
--- a/src/components/Carousel/Carousel.tsx
+++ b/src/components/Carousel/Carousel.tsx
@@ -33,7 +33,7 @@ export interface FlowbiteCarouselIndicatorsTheme {
 
 export interface FlowbiteCarouselItemTheme {
   base: string;
-  wrapper: string;
+  wrapper: FlowbiteBoolean;
 }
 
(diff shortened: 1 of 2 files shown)


src/components/Carousel/Carousel.tsx
```diff
@@ -145,10 +147,10 @@ export const Carousel: FC<CarouselProps> = ({
     >
       <ScrollContainer
         className={twMerge(theme.scrollContainer.base, (isDeviceMobile || !isDragging) && theme.scrollContainer.snap)}
-        draggingClassName="cursor-grab"
+        draggingClassName={draggable ? 'cursor-grab' : ''}
```
rluders, line 150: I think that you can use `null` instead of `''`

rluders (changes requested):

src/components/Carousel/Carousel.tsx
rluders: Actually, I think that you should not change the `draggingClassName` at all, but disable the `vertical` property from the `ScrollContainer`:

https://www.npmjs.com/package/react-indiana-drag-scroll#component-properties

src/components/Carousel/Carousel.tsx
```diff
@@ -145,16 +147,17 @@ export const Carousel: FC<CarouselProps> = ({
...
-        onStartScroll={handleDragging(true)}
+        onStartScroll={handleDragging(draggable)}
         vertical={false}
+        horizontal={draggable}
       >
         {items?.map((item, index) => (
           <div
             key={index}
-            className={theme.item.wrapper}
+            className={theme.item.wrapper + (draggable ? '' : ' cursor-default')}
```
rluders, line 160: You probably need to use `twMerge` function here. Not like the "huge" effect, but it is the default behavior for the components when handling `className` values. Also, I think that could be worth moving this style to the theme. In this case, use the `FlowbiteBoolean`, so you will have something like:

```jsx
className={twMerge(theme.item.wrapper, theme.item.wrapper.draggable[draggable ? 'on' : 'off'])}
```

And you *don't* need to do the same for the `draggingClassName`, just leave has it is. Let me ask @SutuSebastian what his opinion is as well.

rluders (changes requested):

src/components/Carousel/Carousel.tsx
```diff
@@ -145,16 +147,17 @@ export const Carousel: FC<CarouselProps> = ({
     >
       <ScrollContainer
         className={twMerge(theme.scrollContainer.base, (isDeviceMobile || !isDragging) && theme.scrollContainer.snap)}
-        draggingClassName="cursor-grab"
+        draggingClassName={'cursor-grab'}
```
SutuSebastian, line 150: can be reverted to `"cursor-grab"`, no improvement here

src/components/Carousel/Carousel.tsx
Sigabiel: Since I only need to change the cursor-grab flag to cursor-default would it be okay solve that without twMerge but making theme.item.wrapper a Flowbiteboolean entirely.

Theme
```
item: {
    base: 'absolute top-1/2 left-1/2 block w-full -translate-x-1/2 -translate-y-1/2',
    wrapper: {
      off: 'w-full flex-shrink-0 transform cursor-default snap-center',
      on: 'w-full flex-shrink-0 transform cursor-grab snap-center',
    },
  },```

Use
```
<div
            key={index}
            className={theme.item.wrapper[draggable ? 'on' : 'off']}
            data-active={activeItem === index}
            data-testid="carousel-item"
          >
```

src/components/Carousel/Carousel.tsx
Sigabiel: will be included in next commit

rluders (approved):

PR #1151, opened 2023-11-26
fix(pagination): updating lastPage variable correctly for table layout

### LINKED ISSUE #1150: Docs: pagination component issue on documentation page

## Steps to reproduce

1. Go to [this page](https://www.flowbite-react.com/docs/components/pagination#table-data-navigation).
2. Click on the 'Next' button
3. Notice how the page ranges stay the same

## Current behavior

https://github.com/themesberg/flowbite-react/assets/136707592/d00c55d4-a5ac-48ba-8efc-15756ae1a2d2

## Expected behavior

It should go to the next page range (2 to 6)

## Context
Video is recorded using Microsoft Edge, but assuming it wont matter

rajatpandey441:
The page number was not showing correctly on first few clicks because lastPage variable was not updating on the first few clicks when the layout was table due to incorrect addition. Was working correctly when the layout was pagination.

fix #1150

- [ ] I have followed the [Your First Code Contribution section of the Contributing guide](https://github.com/themesberg/flowbite-react/blob/main/CONTRIBUTING.md#your-first-code-contribution)

lastPage variable was set to currentPage+2 which is correct for pagination layout but for table layout, it needs to be currentPage+4 so that lastPage can be 5 after first click only.
#1150 
NA

### DIFF
--- a/src/components/Pagination/Pagination.tsx
+++ b/src/components/Pagination/Pagination.tsx
@@ -65,7 +65,7 @@ const PaginationComponent: FC<PaginationProps> = ({
 }) => {
   const theme = mergeDeep(getTheme().pagination, customTheme);
 
-  const lastPage = Math.min(Math.max(currentPage + 2, 5), totalPages);
+  const lastPage = Math.min(Math.max(layout === 'pagination' ? currentPage + 2 : currentPage + 4, 5), totalPages);
   const firstPage = Math.max(1, lastPage - 4);
 
   const goToNextPage = (): void => {


SutuSebastian (approved):
Looks good!

In the close future, I'd like to make the API even more flexible, and why not even split the logic vs UI, we can make it headless and expose the headless part, maybe people want to do very custom implementation and not re-write pagination logic every time.

rajatpandey441:
I see that this PR is approved. How can I merge this PR? This is my first PR in this repository.

rajatpandey441:
@SutuSebastian Thanks for letting me know. Appreciate it.

tulup-conner (approved):
Agree with @SutuSebastian, this is great for now but we should definitely look to provide the user with direct access to how this works instead of doing it for them. The current API is confusing in this regard.

rajatpandey441:
We can also add a attribute "pageSize" rather than defining 5 records per page for better flexibility.

Issue #871, opened 2023-07-20
Checkbox: add 'indeterminate' prop


tulup-conner:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

### Discussed in https://github.com/themesberg/flowbite-react/discussions/870

<div type='discussions-op-text'>

<sup>Originally posted by **grasdira** July 19, 2023</sup>
#### Enhancement Suggestion
I think it would be nice to add `indeterminate` (or `isIndeterminate`) prop to Checkbox component.
</div>

Example:

```js
<Checkbox indeterminate .. />
```

## Context

This is officially supported by DOM elements in all browsers. See https://developer.mozilla.org/en-US/docs/Web/CSS/:indeterminate

marcadrian-it:
Hey, could I take this?

grasdira:
Hi! I'd like to take a shot at solving this issue. Could you assign it to me? Thanks

tulup-conner:
@grasdira Done, but for the future, you are always welcome to just get started on an issue!

shereen-fathy:
Hi @grasdira, any progress on this issue?

grasdira:
Hi @shereen-fathy, yes, I've made some progress on this issue.

azerazerty:
is this fixed or not yet ?

SutuSebastian:
Fixed by #1498  in [flowbite-react@0.11.0](https://github.com/themesberg/flowbite-react/releases/tag/flowbite-react%400.11.0) release.

PR #1155, opened 2023-11-28
feat(checkbox): add checkbox indeterminate prop

grasdira:

Start from the discussion #870 

Related to #871 

There are no breaking API changes. 
Here are my changes:

1. To ensure that checkboxes with an indeterminate state display the correct style, this commit(https://github.com/themesberg/flowbite-react/commit/65e9f8bbfeb395999ccfd63cd21c606a2a181a5d) added the following CSS in both `.storybook/style.css` and `docs.css` to override the default styles:
``` css
[type='checkbox']:indeterminate {
  background-image: url('data:image/svg+xml;base64,PHN2ZyBhcmlhLWhpZGRlbj0ndHJ1ZScgeG1sbnM9J2h0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnJyBmaWxsPSdub25lJyB2aWV3Qm94PScwIDAgMTYgMTInPgogIDxwYXRoIHN0cm9rZT0nd2hpdGUnIHN0cm9rZS1saW5lY2FwPSdyb3VuZCcgc3Ryb2tlLXdpZHRoPSczJyBkPSdNMCA2SDE2Jy8+Cjwvc3ZnPg==');
  background-color: currentColor;
}
```

2. This commit(https://github.com/themesberg/flowbite-react/commit/f1a847874af3ce786e3d04fd189273f146e31276) added a new example of using the indeterminate state to handle multiple checkboxes in a group or tree structure.

### DIFF
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -1,5 +1,7 @@
+'use client';
+
 import type { ComponentProps } from 'react';
-import { forwardRef } from 'react';
+import { forwardRef, useEffect, useRef } from 'react';
 import { twMerge } from 'tailwind-merge';
 import { mergeDeep } from '../../helpers/merge-deep';
 import { getTheme } from '../../theme-store';
@@ -17,15 +19,33 @@ export interface FlowbiteCheckboxRootTheme {
 export interface CheckboxProps extends Omit<ComponentProps<'input'>, 'type' | 'ref' | 'color'> {
   theme?: DeepPartial<FlowbiteCheckboxTheme>;
   color?: keyof FlowbiteColors;
+  indeterminate?: boolean;
 }
 
 export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
-  ({ className, color = 'default', theme: customTheme = {}, ...props }, ref) => {
+  ({ className, color = 'default', theme: customTheme = {}, indeterminate, ...props }, ref) => {
     const theme = mergeDeep(getTheme().checkbox, customTheme);
+
+    const inputRef = useRef<HTMLInputElement | null>(null);
+
+    useEffect(() => {
+      if (inputRef.current && indeterminate !== inputRef.current.indeterminate) {
+        inputRef.current.indeterminate = !!indeterminate;
+      }
+    }, [indeterminate]);
 
     return (
       <input
-        ref={ref}
+        ref={(element) => {
+          inputRef.current = element;
+          if (ref) {
+            if (typeof ref === 'function') {
+              ref(element);
+            } else {
+              ref.current = element;
+            }
+          }
+        }}
         type="checkbox"
         className={twMerge(theme.root.base, theme.root.color[color], className)}
         {...props}
--- a/src/components/Checkbox/Checkbox.tsx
+++ b/src/components/Checkbox/Checkbox.tsx
@@ -47,12 +47,7 @@ export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
           }
         }}
         type="checkbox"
-        className={twMerge(
-          'indeterminate:bg-indeterminate dark:indeterminate:bg-current dark:indeterminate:border-gray-600',
-          theme.root.base,
-          theme.root.color[color],
-          className,
-        )}
+        className={twMerge(theme.root.base, theme.root.color[color], className)}
         {...props}
       />
     );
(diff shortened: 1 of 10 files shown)


tulup-conner:
Hey thanks for the contribution! I don't have the time to look at this right now but forgive me for jumping in and rebasing, we're trying to figure out what's wrong with codecov. Thank you for your patience! @grasdira

src/components/Checkbox/Checkbox.tsx
```diff
@@ -17,15 +19,33 @@ export interface FlowbiteCheckboxRootTheme {
...
   color?: keyof FlowbiteColors;
+  indeterminate?: boolean;
 }
 
 export const Checkbox = forwardRef<HTMLInputElement, CheckboxProps>(
-  ({ className, color = 'default', theme: customTheme = {}, ...props }, ref) => {
+  ({ className, color = 'default', theme: customTheme = {}, indeterminate, ...props }, ref) => {
     const theme = mergeDeep(getTheme().checkbox, customTheme);
 
+    const inputRef = useRef<HTMLInputElement | null>(null);
```
SutuSebastian, line 29: can't this be done using custom data attribute prop (eg: `data-...`) instead of injecting it through ref and checking the data attribute with tailwind directly in this file? 🤔

SutuSebastian:
> Anyone still working on this?

Doesn't seem like it 🤔

grasdira:
I'm really sorry for not keeping up with this PR 😔
Since not quite familiar with this project, I need more help to discuss how to solve this problem.

grasdira:
Hi! 
After rethinking the code, I have made some adjustments to provide developers with more flexibility when using the component:

1. I am continuing to use `useRef` to directly set the indeterminate property on the DOM element. This enables the use of the pseudo-class modifiers `indeterminate:` in TailwindCSS for CSS customization. (@SutuSebastian sorry for taking so long to reply, this point is why I don't think `data-...` is a better way on this feature. )
2. I have removed the modifications I previously made in two `.css` files, realizing that this approach was limiting the developers’ ability to change the icon or other styles for the indeterminate state. (https://github.com/themesberg/flowbite-react/pull/1155/commits/90ec304ff705e76d7bb30d729df1c409f450076f)
3. In `tailwind.config.ts`, I have extended indeterminate `backgroundImage`(https://github.com/themesberg/flowbite-react/pull/1155/commits/a4af18f22c8735ea21630319064efc6bee1df8d0)  and set the CSS for the indeterminate checkbox with `indeterminate:` modifier in `Checkbox.tsx`(https://github.com/themesberg/flowbite-react/pull/1155/commits/a1bfc2134cbfb233f51b49cf5372bf788973ceef).

What are your opinions on these changes?

src/components/Checkbox/Checkbox.tsx
grasdira: Hi, @SutuSebastian . I have provided my responses in this [comment](https://github.com/themesberg/flowbite-react/pull/1155#issuecomment-1896256180). Please review them at your earliest convenienceI.

tailwind.config.ts
SutuSebastian, line 1: Enforcing the users to apply the same config into their tailwind config is not the way to go, the only thing they'd have to configure is primary and secondary colors override, if needed.

On top of that, needing a custom tailwind config just for a simple component state, is a bit over-engineered. We need to find a simple solution here, that does not spill out more than the components context/folder.

grasdira: I certainly understand we should not ask user to update their Tailwind config. But I do noticed that the indeterminate state's style definition isn't covered by default in Flowbite. 

Is it a better way to integrate indeterminate state's styles directly into Flowbite's default stylesheet, so that all users have access without extra configuration?

SutuSebastian: U mean to include the indeterminate into flowbite/plugin?

I'd rather have it at the component level, its just a matter of a dash icon and an extra prop.

SutuSebastian: I can also see in the core library plugin, we already have it: https://github.com/themesberg/flowbite/blob/main/plugin.js#L350

grasdira: I wasn't aware that the indeterminate state's style had already been defined in the plugin. Thank you for pointing me to it.
I'll update and resubmit shortly.

grasdira: @SutuSebastian Hey, am I allowed to update `flowbite` to the latest version. 
The indeterminate state icon changed in a recent [commit](https://github.com/themesberg/flowbite/commit/ec0eab5005383cba80fa1664353897d74d7ef7b1) , and it's causing the icon not to change in dev mode. It used to be a check icon :astonished:

SutuSebastian: Sure, go ahead

Issue #1159, opened 2023-11-30
`<Checkbox>` doesn't have outline in dark mode


tulup-conner:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Steps to reproduce

1. Create any `<Checkbox />`
2. Switch to dark mode
3. Click on it

## Current behavior

The checkbox is surrounded by a thick layer of blue.

## Expected behavior

There should be a small (~2px) gap between the edge of the checkbox itself and the blue border around it. I believe the box model term for this space is the outline.

## Context

Reproducible in the live docs: https://www.flowbite-react.com/docs/components/forms#checkbox
Correct behavior: https://flowbite.com/docs/forms/checkbox/#checkbox-example

gramzanali-scottlogic:
Hi, I'm looking to make a first contribution to an open source project. Can I have a go at fixing this issue?

mud1tx:
@tulup-conner I can give this issue a try. Can you please assign it to me.

mud1tx:
@tulup-conner earlier it looks like this
![Screenshot from 2023-12-01 11-34-25](https://github.com/themesberg/flowbite-react/assets/89539376/80008a61-e42f-403d-bd05-508ec220473e)

now it is looking like this
![Screenshot from 2023-12-01 11-32-57](https://github.com/themesberg/flowbite-react/assets/89539376/c2ceab4c-8d10-402e-8048-f52d28eae018)

is this works or I have to do some changes?

tulup-conner:
That looks right at first glance! You should check out the official vanilla Flowbite library docs to check that anything you're working on. For Checkboxes, that's https://flowbite.com/docs/forms/checkbox/#checkbox-example

PR #1160, opened 2023-12-01
Fix : CheckBox Dark Mode Bug

mud1tx:

- Just change one css line in theme.ts

- #1159 


tulup-conner (approved):
Thanks so much! 

BTW, it looks like this is actually an issue for Radios in dark mode as well. If you have time to take that on it would also be much appreciated. It looks like that one's a bit different - the ring is white and also cyan.

PR #1166, opened 2023-12-03
feat(spinner): add missing colors to spinner component

MateoWartelle:

- Added missing colors that extend from FlowbiteColors
-  The issue I currently see with (Spinner, Badge and Button). The colors info failure, success, warning, dark, light are kind of weird to have on all components if the colors are going to be duplicated. 


tulup-conner (changes requested):
Thanks for the addition of the missing colors, but I can't merge this with the exclusion of the semantic colors. It's a format we have for every other component. What we should do is add all the colors by their color name, then additionally point semantic colors to those. For example, we should have both danger and red, and assign danger to red's value.

PR #1173, opened 2023-12-06
fix(datepicker): fix the inconsistent weekday labels

### LINKED ISSUE #1044: Inconsistent Datepicker Weekday Labels

## Steps to reproduce

1. Go to https://www.flowbite-react.com/docs/components/datepicker
2. Click on the first datepicker example to view selector

## Current behavior

The datepicker component is currently showing different weekdays for different users.

User A:
image

User B:
image

## Expected behavior

Ideally these are always correct regardless of timezone, locale, or browser.

In this case, weekday labels should be starting on Sunday, and ending on Saturday to match correctly with the numbers below.

## Context

In my case I fall into the "User A" category, seeing the wrong weekday labels. Can replicate this on Edge and Brave Browser (both Chromium-based). I am currently on GMT-4 timezone-wise. English locale.

This could be caused by a variety of issues as hinted above - the default `weekStart` [on the component](https://github.com/themesberg/flowbite-react/blob/main/src/components/Datepicker/Datepicker.tsx#L102) is set to Sunday, and I don't see this being overwritten on the docs app, so maybe something is wrong with how context is being gathered [here](https://github.com/themesberg/flowbite-react/blob/main/src/components/Datepicker/DatepickerContext.tsx#L23), either due to locale or timezone specifics.

Alternatively, this could be due to timezone or locale issues on the `getWeekDays()` function [here](https://github.com/themesberg/flowbite-react/blob/main/src/components/Datepicker/helpers.ts#L66), although somewhat unlikely.

markhker:
Updating the WeekStart enum to be aligned with the values Javascript Date object uses allows to simplify the code and maintain compaitbility. Fixed the off-by-one error in the weekdays order.

BREAKING CHANGE: As the WeekStart enum changed order to be aligned with Javascript Date object, now you have to change your weekStart attribute to be -1, so for Monday you should put 1, instead of 2. But it wasn't working before as the values were incorrectly rendered in the first place.

fix #1044

#1044

As the WeekStart enum changed order to be aligned with Javascript Date object, now you have to change your weekStart attribute to be -1, so for Monday you should put 1, instead of 2. But it wasn't working before as the values were incorrectly rendered in the first place.

### DIFF
--- a/content/docs/components/datepicker.mdx
+++ b/content/docs/components/datepicker.mdx
@@ -39,13 +39,13 @@ The `weekStart` prop can be used to set the first day of the week inside the dat
 
 ```json
 {
-  "0": "Saturday",
-  "1": "Sunday",
-  "2": "Monday",
-  "3": "Tuesday",
-  "4": "Wednesday",
-  "5": "Thursday",
-  "6": "Friday"
+  "0": "Sunday",
+  "1": "Monday",
+  "2": "Tuesday",
+  "3": "Wednesday",
+  "4": "Thursday",
+  "5": "Friday",
+  "6": "Saturday"
 }
 ```
 
--- a/examples/datepicker/datepicker.weekStart.tsx
+++ b/examples/datepicker/datepicker.weekStart.tsx
@@ -9,7 +9,7 @@ import { Datepicker } from 'flowbite-react';
 function Component() {
   return (
     <Datepicker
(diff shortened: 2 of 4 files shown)


tulup-conner:
@rluders Can you check this out?

SutuSebastian (approved):
LGTM

PR #1194, opened 2023-12-18
docs: new integration guides

SutuSebastian:

### Summary

Integration guides for all current React related frameworks/technologies that `flowbite-react` can be installed into.

### Changes

- [x] replace `@mapbox/rehype-prism` in favor of `rehype-prism-plus` for line highlighting features (see screenshots below)
- [x] create new "integration guides" section
- [x] create new integration guides for the following:
  - [x] Next.js 
  - [x] Vite 
  - [x] Remix 
  - [x] Astro 
  - [x] Gatsby
  - [x] Laravel 
  - [x] Parcel 
  - [x] Create React App 
  - [x] RedwoodJS
- [x] **explicit** dark mode support integration for the frameworks that do SSR
- [x] use `Link` from Next.js for internal routes

### Result

New "Integration guides" section

Screenshot 2023-12-18 at 13 38 02

New code highlight line

Screenshot 2023-12-18 at 13 38 54


content/docs/guides/gatsby.mdx
```diff
@@ -0,0 +1,82 @@
...
+```
+
+2. Select `"Tailwind CSS"` to `"Would you like to install a styling system?"` question.
+
+## Install Flowbite React
+
+1. Run the following command to install `flowbite` and `flowbite-react`:
+
+```bash
+npm i flowbite flowbite-react
```
tulup-conner, line 21: I believe we can safely remove the `flowbite` part of `npm i` for all of these guides, because it is an implicit dependency of `flowbite-react`

content/docs/guides/next-js.mdx
```diff
@@ -0,0 +1,180 @@
...
+```
+
+3. Add the paths to all of your template files in your `tailwind.config.js` file:
+
+```js {4-9}
+/** @type {import('tailwindcss').Config} */
+module.exports = {
+  content: [
+    './app/**/*.{js,ts,jsx,tsx,mdx}',
+    './pages/**/*.{js,ts,jsx,tsx,mdx}',
```
tulup-conner, line 69: You'll want to make a comment callout here differentiating Next 13 and <= 12. Like the one for the `src` directory, but

- If using >= 13 Next App Router, `app`
- If using <= 12 Next Pages Router, `pages` and `components`

package.json
```diff
@@ -62,12 +62,12 @@
     "react-icons": "^4.11.0",
     "react-indiana-drag-scroll": "^2.2.0",
     "react-markdown": "^9.0.0",
+    "rehype-prism-plus": "^1.6.3",
```
tulup-conner, line 65: This is another instance of a package we are including in the library now because it needs to be a production dependency for Next.js. Is there a way to tell Vercel to install `devDependencies` when it builds the website? That would help us get around needing to have these as production dependencies, which are included for every user who runs `npm i flowbite-react` (until monorepo)

Also, I haven't actually tried it but in this instance, you may well be able to install this as a development dependency and it works just fine, since the original rehype plugin did. Should probably try that first.

tulup-conner (changes requested):
I don't have time to check the Astro, CRA, Laravel, RedwoodJS instructions but your instructions for the rest, which I have used a few times myself, look great.

SutuSebastian:
> I don't have time to check the Astro, CRA, Laravel, RedwoodJS instructions but your instructions for the rest, which I have used a few times myself, look great.

Every single guide was tested by hand, from scratch, so I think we're covered there :D

package.json
SutuSebastian: I think we should not worry about extra dependencies ending up on users since the monorepo is just around the corner, which invalidates all this at times.

SutuSebastian: I will go through every single thing once the monorepo work is started and I think that's the best time to sort all packages installed and put them right where they should be (dep, or devDep)

content/docs/guides/next-js.mdx
SutuSebastian: This is coming from the official documentation of tailwind (see https://tailwindcss.com/docs/guides/nextjs), I'd rather keep all of the cases at once, or I can separate them into 2 code blocks (1 for app router and 1 pages router), but I think that might bloat the content with more and more instructions.

Keeping it as vanilla as possible was the goal here (with official TW docs)

content/docs/guides/gatsby.mdx
SutuSebastian: agreed

content/docs/guides/next-js.mdx
SutuSebastian: Also, we can take into account that some big scale projects are on the process of migrating their app from `pages` router to `app` router and they have both enabled until migration is complete, therefore this setup caters for both cases.

package.json
tulup-conner: Yeah I just wanted to keep that on your radar. It's already annoying people (rightly so) and will continue to do so

content/docs/guides/next-js.mdx
tulup-conner: Good callout, that makes sense!

tulup-conner (approved):

package.json
SutuSebastian: Totally agreed!

Issue #589, opened 2023-02-10
Feature request: tooltip delay


multiwebinc:
As it is, tooltips show up (or begin to show up) immediately on mouse over. I propose adding a `delay` attribute to the `Tooltip` component that will delay showing by that many milliseconds.

robbyevans:
hello, @rluders @multiwebinc, I  would love to solve this issue.

tulup-conner:
@robbyevans Are you still able to resolve this?

robbyevans:
@tulup-conner yes I am available

jam3smay:
I'd also love to see this solved with a delay prop or something like trigger="hover:delay-500". Without the ability to configure a delay, this component is useless for my use cases at least. Thnx.

anuj8126:
Hi, I’d like to work on this if it’s still available.
I’ll first check whether the Tooltip delay can be supported cleanly through the existing theme API or whether a small `delay` prop is needed. I’ll keep the PR focused and include a docs/story example if required.

PR #1200, opened 2023-12-23
feat(tooltip): adds tooltip support for restMs and delay properties

DavidSeptimus:
implements #589

This change adds support for floating-ui's `useHover` [`restMs`](https://floating-ui.com/docs/usehover#restMs) and [`delay`](https://floating-ui.com/docs/usehover#delay) props to the `Tooltip` component to enable delayed opening/closing of tooltips.

### DIFF
--- a/content/docs/components/tooltip.mdx
+++ b/content/docs/components/tooltip.mdx
@@ -53,6 +53,16 @@ You can disable the arrow of the tooltip component by passing the `arrow` prop w
 
 <Example name="tooltip.disableArrow" />
 
+## Hover delay
+
+When your tooltip uses the `hover` trigger type, use the `restMs` prop to specify the amount of time (in milliseconds) the user's cursor must be "resting" on the trigger before the tooltip appears.
+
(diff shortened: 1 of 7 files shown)


tulup-conner:
I understand the practical appeal, but this functionality isn't available in the upstream library at flowbite.com, so I'm not sure if we should add it here. 

@zoltanszogyenyi I defer to you on this one - it seems like this would be a useful feature, and it's also pretty difficult to create a wrapper of our component to achieve this. You would probably be forced to patch flowbite-react to do so.

DavidSeptimus-Klotho:
It looks like the upstream has support for `trigger="none"` and exposes `show()`, `hide()`, and `toggle()` functions for custom triggers. It just seems kind of wasteful to have to reimplement  delay when it's supported by the underlying floating-ui implementation.

SutuSebastian:
Nice one! It only needs re-build since we now have new monorepo structure (see #1264).

If u feel like u don't have the time, I can take it from here, just let me know! 🤝

DavidSeptimus-Klotho:
@SutuSebastian I doubt I'll have time for this in the near-term. Feel free to pick it up if you'd like.

PR #1207, opened 2023-12-31
feat(carousel): automatic hiding of navigation buttons when the first…

### LINKED ISSUE #1205: Carousel auto hide controls

## Summary

Hello. I came across the necessity to implement carousel controls hiding when the first or last image is active. I know this is not supported in native flowbite carousel but it gives a nice touch for UX when needed. 
I implemented this solution in my forked repo. Added an optional parameter to carousel component with false default value.

https://github.com/themesberg/flowbite-react/assets/36115659/be0b1ec4-6ebe-4354-8a44-00777ead9d42

## Context

I will create a PR in case you would like to add the feature to this repo.

cristiantiradob:
feat(carousel): automatic hiding of navigation buttons when the first or last image is active

Added 'autoHideControls' feature to ImageSlider for intuitive navigation button visibility.

Gives a nice touch for UX when not needing to move from last image to first one.
Conditional rendering of carousel controls when autoHideControls is true.

https://github.com/themesberg/flowbite-react/assets/36115659/87521d58-b0b4-4d34-bd78-d59b950d10a7

fixes #1205 

New optional prop added to Carousel component: autoHideControls with false as default value.
Example:
Captura de pantalla 2023-12-31 a la(s) 4 34 23 p m

### DIFF
--- a/src/components/Carousel/Carousel.tsx
+++ b/src/components/Carousel/Carousel.tsx
@@ -56,6 +56,7 @@ export interface CarouselProps extends ComponentProps<'div'> {
   theme?: DeepPartial<FlowbiteCarouselTheme>;
   onSlideChange?: (slide: number) => void;
   pauseOnHover?: boolean;
+  autoHideControls?: boolean;
 }
 
 export interface DefaultLeftRightControlProps extends ComponentProps<'div'> {
@@ -74,6 +75,7 @@ export const Carousel: FC<CarouselProps> = ({
   theme: customTheme = {},
   onSlideChange = null,
   pauseOnHover = false,
+  autoHideControls = false,
   ...props
 }) => {
   const theme = mergeDeep(getTheme().carousel, customTheme);
@@ -181,7 +183,11 @@ export const Carousel: FC<CarouselProps> = ({
 
       {items && (
         <>
-          <div className={theme.root.leftControl}>
+          <div
+            className={`${theme.root.leftControl} transition-opacity duration-300 ${
+              activeItem === 0 && autoHideControls ? 'pointer-events-none opacity-0' : 'opacity-100'
+            }`}
+          >
             <button
               className="group"
               data-testid="carousel-left-control"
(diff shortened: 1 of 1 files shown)


tulup-conner:
I like this idea and solution a lot, but like you said, it's not in the vanilla library so we probably shouldn't add it. @zoltanszogyenyi What do you think here? This PR looks good if we want it.

src/components/Carousel/Carousel.tsx
```diff
@@ -192,7 +198,11 @@ export const Carousel: FC<CarouselProps> = ({
               {leftControl ? leftControl : <DefaultLeftControl theme={customTheme} />}
             </button>
           </div>
-          <div className={theme.root.rightControl}>
+          <div
+            className={`${theme.root.rightControl} transition-opacity duration-300 ${
+              activeItem === items.length - 1 && autoHideControls ? 'pointer-events-none opacity-0' : 'opacity-100'
+            }`}
+          >
```
ddiasfront, line 205: Thanks for your contribution.

Would opacity be the best approach in this case? W/opacity the controls will remain available to user interactions on the dom, while display:none would prevent, that.

tulup-conner: Good catch, yes, that's right.

cristiantiradob: Hello. If with controls remaining available you mean that the user will go from last image to first image even with control hided, in my tests when control hides click has no effect. In this video you will see a circle in the pointer when I click:

https://github.com/themesberg/flowbite-react/assets/36115659/49c60eb2-3795-4482-887b-2dfbb117bba5

Please correct me if I misunderstood.

tulup-conner: I would still change this to `display:none` because this may come up as still clickable for users navigating with Tab.

I can't confirm that since you didn't add an example to the docs, which, if you could, would be great.

With that said, I'm still not sure if we can merge this because this is a UX feature that isn't available in the vanilla library at all.

cristiantiradob: You are right! Thanks for your feedback I will implement this in my forked repo. Please let me know if you'd like this feature here so I can update the PR.

SutuSebastian: Keep in mind that if the navigation button/icon disappears from the DOM, when the user spam clicks next (or prev) until he gets to the last, if in the future we'll implement click-to-fullscreen (or similar), then this behaviour would be very unpredictable.

So the case study is, the next/prev icons are wrapped in a `button` element in order to be focusable/accessible and the button becomes `disabled=true` when is last or first page, according to the direction, this way tabbing is not possible (CC @tulup-conner ) and also clicking the disabled button should stop event propagation.

#just_thoughts 🤔

PR #1214, opened 2024-01-04
Feature/carousel active slide

### LINKED ISSUE #1094: Setting Active Image In the Slider
We can programatically set the active image: https://flowbite.com/docs/components/carousel/#default-slider

But the same is not available in flowbite react: https://www.flowbite-react.com/docs/components/carousel

Can we please get this in React too?

**Expected Behaviour:** 
When we set `data-carousel-item="active"` or something similar on any image, that image should be shown in the carousel and not from index 0.

cristiantiradob:

Added 'activeSlide' prop with tests and docs. I considered the following facts:

- Validate 'activeSlide' prop:
  If negative value, default slider item is set to '0'. If value is greater than carousel childs, default slider item is set to the last one. If no carousel childs, exit useEffect.

- Default value:
  If no 'activeSlide' prop provided, default value is set to 0.

- Update 'activeItem' State
  'activeItem' is set to 'activeSlide' when the component mounts.

Fixes #1094

### DIFF
--- a/content/docs/components/carousel.mdx
+++ b/content/docs/components/carousel.mdx
@@ -31,6 +31,12 @@ Use the `slideInterval` prop to set the interval between slides in milliseconds.
 
 <Example name="carousel.slideInterval" />
 
+## Active Slide item
+
+Use `activeSlide` prop to set the default active slider item based on its position (starts from [0-x]). Default value is `0`.
+
+<Example name="carousel.activeSlide" />
+
 ## Custom controls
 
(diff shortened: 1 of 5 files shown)


shippinglob:
Hi. I'm not one of the maintainer of this repository but thanks for implementing this very useful update. I went over the PR and It LGTM. Only a minor comment for :

`navigateTo(validateActiveSlide)(); `

In the case of an invalid index, I prefer keeping the current active item as it makes debugging issues easier.

PR #1222, opened 2024-01-09
Fix/range slider align

### LINKED ISSUE #1202: Range slider is not vertically centered

## Steps to reproduce

1. Go to the [Flowbite React official doc](https://www.flowbite-react.com/docs/components/forms#range-slider)
2. With dev tools, highlight the surrounding div of the first range slider:
image
3. Look at the range slider container. The range is not vertically centered in the container. This is clearly visible with the blue circle: it touches the bottom of the container box but not the top.
image

## Current behavior

The ranger slider is not centered in its container. Therefore it is hard to vertically center the whole component.

## Expected behavior

The input fields should be vertically centered inside the range slider container.

## Context

This behavior can prevent a range slider from being vertically centered along with other components, for example with a `flex` and `items-center`. The range slider is "a little lower" than the other vertically centered components.

This behavior can be observed with Chrome 120.0.6099.130 and Firefox 121.0. The issue can be observed right from the Flowbite React docs, so it must be present in the latest Flowbite React version.

revuwem:

Removed field base wrapper from RangeSlider according to vanilla [Flowbite range slider](https://flowbite.com/docs/forms/range/#range-slider-example) implementation to fix issue with component vertical alignment

Fixes #1202

Breaking API changes:  `className` props now applies to input instead of component field base wrapper.


src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -34,16 +34,12 @@ export const RangeSlider = forwardRef<HTMLInputElement, RangeSliderProps>(
...
-            <input
-              ref={ref}
-              type="range"
-              className={twMerge(theme.field.input.base, theme.field.input.sizes[sizing])}
-              {...props}
-            />
-          </div>
-        </div>
+        <input
+          ref={ref}
```
tulup-conner, line 38: Do we need [forwardRef](https://react.dev/reference/react/forwardRef) for this also?

src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -34,16 +34,12 @@ export const RangeSlider = forwardRef<HTMLInputElement, RangeSliderProps>(
...
+          type="range"
+          className={twMerge(theme.field.input.base, theme.field.input.sizes[sizing], className)}
```
tulup-conner, line 40: Since you're doing this, it would be helpful to consolidate the theme API here since we're making the component much simpler, as it should be. It seems like you could change the theme to literally just `base` and `sizes` keys. 

Remember to consult [the Flowbite vanilla docs on RangeSlider](https://flowbite.com/docs/forms/range/#range-slider-example) for the classes in the theme. I think we have a mismatch already, before your PR.

tulup-conner (changes requested):
Thank you!

src/components/RangeSlider/RangeSlider.tsx
revuwem: @tulup-conner I didn't quite understand your question. I believe, if we want `<RangeSlider />` to support refs, we do, don't we? Or is this about whether we need this functionality in the first place?

src/components/RangeSlider/RangeSlider.tsx
revuwem: @tulup-conner Good point, thanks!

src/components/RangeSlider/RangeSlider.tsx
tulup-conner: No that's my fault, I only looked at the diff and didn't notice this already has forwardRef

src/components/RangeSlider/RangeSlider.tsx
```diff
@@ -34,16 +22,7 @@ export const RangeSlider = forwardRef<HTMLInputElement, RangeSliderProps>(
 
     return (
       <>
```
SutuSebastian, line 24: in this case, do we still need the fragment (`<>`)?

src/components/RangeSlider/RangeSlider.spec.tsx
```diff
@@ -5,7 +5,7 @@ import { describe, expect, it, vi } from 'vitest';
 import { Flowbite, type CustomFlowbiteTheme } from '../Flowbite';
 import { RangeSlider } from './RangeSlider';
 
-describe('Components / Button', () => {
+describe('Components / RangeSlider', () => {
```
SutuSebastian, line 8: good catch & cleanup

PR #1224, opened 2024-01-09
Fix/timeline vertical

### LINKED ISSUE #1196: TimelineContent is not taking theme styles when being in vertical mode

## Steps to reproduce

1. Use the Timeline.Content component and configuring the theme to change the styling
2. The style is applied only when horizontal prop is "true"

## Current behavior

Currently, the Timeline.Content component, when passing the styles from the theme, is not taking properly the styles because they are only applied in horizontal.
image

Example: 
This is an example of the same theme used on horizontal and in vertical

This is my theme object
image

Horizontal:
image

Vertical:
Captura de pantalla 2023-12-21 a las 17 09 45

As you can see, the border from the box is only applied when horizontal prop is true

## Expected behavior

Both horizontal and vertical options apply the root base from content

## Context

Right now, when using the Timeline component, you can add a child which is Timeline.Content which is basically the content inside each item, but this is only letting change the style when using it in horizontal.

revuwem:

Fixed the problem with vertical-oriented `<Timeline.Content />` component style. The `root.base` class theme prop is now independent on orientation. The optional `root.vertical` and `root.horizontal` props were introduced to make orientation-specific style changes.

Fix #1196

`interface FlowbiteTimelineContentTheme` has been changed. The theme now supports optional `root.vertical` and `root.horizontal` props.

### DIFF
--- a/src/components/Timeline/TimelineContent.tsx
+++ b/src/components/Timeline/TimelineContent.tsx
@@ -14,6 +14,8 @@ import type { FlowbiteTimelineTitleTheme } from './TimelineTitle';
 export interface FlowbiteTimelineContentTheme {
   root: {
     base: string;
+    vertical?: string;
+    horizontal?: string;
   };
   time: FlowbiteTimelineTitleTheme;
   title: FlowbiteTimelineTimeTheme;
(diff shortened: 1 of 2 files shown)


SutuSebastian (approved):
looks good

src/components/Timeline/TimelineContent.tsx
```diff
@@ -14,6 +14,8 @@ import type { FlowbiteTimelineTitleTheme } from './TimelineTitle';
 export interface FlowbiteTimelineContentTheme {
   root: {
     base: string;
+    vertical?: string;
```
SutuSebastian, line 17: these are redundant, can be removed

src/components/Timeline/TimelineContent.tsx
```diff
@@ -14,6 +14,8 @@ import type { FlowbiteTimelineTitleTheme } from './TimelineTitle';
...
+    horizontal?: string;
```
SutuSebastian, line 18: same here

SutuSebastian (changes requested):
looks good overall, just those 2 comments needs fixing

SutuSebastian:
ported to #1430 due to new monorepo structure

SutuSebastian:
ported to #1430 due to new monorepo structure

PR #1238, opened 2024-01-21
fix(button): Button ring on focus

nigellima:
fixes the button inconsistent focus ring compared to the core library. It also adds a transition of colors and ring

This PR fixes the button focus ring, which is currently inconsistent with the core library. The Button Group ring was wrongly overriding the ring-4 classes defined in the button themes.

Also, I took the opportunity to add a slight transition on the button colors/shadows to have the hover and the ring styles animated

button-ring


SutuSebastian (approved):
ggwp

PR #1247, opened 2024-01-24
fix(navbar): on Mobile Screen Navbar remains open even page is changed

dhavalveera:
When we go to the Mobile Screen and when we click on Navbar, it opens the Collapse, and when we click on any of the Link and it get redirected to another page, but Navbar Collapse isn't getting closed, it was only getting closed from Navbar Toggle nowhere else.

This fix is adding the onClick event to NavbarCollapse so whenever we click on the Link, it is closing the navbar.

**Current:**
image

**Fixed:**
image

### DIFF
--- a/src/components/Navbar/NavbarCollapse.tsx
+++ b/src/components/Navbar/NavbarCollapse.tsx
@@ -18,22 +18,14 @@ export interface NavbarCollapseProps extends ComponentProps<'div'> {
 }
 
 export const NavbarCollapse: FC<NavbarCollapseProps> = ({ children, className, theme: customTheme = {}, ...props }) => {
-  const { theme: rootTheme, isOpen, setIsOpen } = useNavbarContext();
+  const { theme: rootTheme, isOpen } = useNavbarContext();
 
   const theme = mergeDeep(rootTheme.collapse, customTheme);
-
-  const handleClick = () => {
-    setIsOpen(!isOpen);
-  };
 
   return (
     <div
       data-testid="flowbite-navbar-collapse"
       className={twMerge(theme.base, theme.hidden[!isOpen ? 'on' : 'off'], className)}
-      onClick={handleClick}
-      onKeyDown={handleClick}
-      role="button"
-      tabIndex={0}
       {...props}
     >
       <ul className={theme.list}>{children}</ul>
--- a/src/components/Navbar/NavbarCollapse.tsx
+++ b/src/components/Navbar/NavbarCollapse.tsx
@@ -18,14 +18,19 @@ export interface NavbarCollapseProps extends ComponentProps<'div'> {
 }
 
 export const NavbarCollapse: FC<NavbarCollapseProps> = ({ children, className, theme: customTheme = {}, ...props }) => {
-  const { theme: rootTheme, isOpen } = useNavbarContext();
+  const { theme: rootTheme, isOpen, setIsOpen } = useNavbarContext();
 
   const theme = mergeDeep(rootTheme.collapse, customTheme);
+
+  const handleClick = () => {
+    setIsOpen(!isOpen);
+  };
 
   return (
     <div
       data-testid="flowbite-navbar-collapse"
       className={twMerge(theme.base, theme.hidden[!isOpen ? 'on' : 'off'], className)}
+      onClick={handleClick}
       {...props}
     >
       <ul className={theme.list}>{children}</ul>
--- a/src/components/Navbar/NavbarCollapse.tsx
+++ b/src/components/Navbar/NavbarCollapse.tsx
@@ -31,6 +31,9 @@ export const NavbarCollapse: FC<NavbarCollapseProps> = ({ children, className, t
       data-testid="flowbite-navbar-collapse"
       className={twMerge(theme.base, theme.hidden[!isOpen ? 'on' : 'off'], className)}
       onClick={handleClick}
+      onKeyDown={handleClick}
+      role="button"
+      tabIndex={0}
       {...props}
     >
       <ul className={theme.list}>{children}</ul>
--- a/src/components/Navbar/NavbarLink.tsx
+++ b/src/components/Navbar/NavbarLink.tsx
@@ -31,13 +31,13 @@ export const NavbarLink: FC<NavbarLinkProps> = ({
   onClick,
   ...props
 }) => {
-  const { theme: rootTheme, isOpen, setIsOpen } = useNavbarContext();
+  const { theme: rootTheme, setIsOpen } = useNavbarContext();
 
   const theme = mergeDeep(rootTheme.link, customTheme);
 
   const handleClick = (event: MouseEvent<HTMLAnchorElement>) => {
-    setIsOpen(!isOpen);
-    onClick && onClick(event);
+    setIsOpen(false);
+    onClick?.(event);
   };
 
   return (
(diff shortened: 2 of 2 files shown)


src/components/Navbar/NavbarCollapse.tsx
```diff
@@ -18,14 +18,22 @@ export interface NavbarCollapseProps extends ComponentProps<'div'> {
...
+    setIsOpen(!isOpen);
+  };
+
   return (
     <div
       data-testid="flowbite-navbar-collapse"
       className={twMerge(theme.base, theme.hidden[!isOpen ? 'on' : 'off'], className)}
+      onClick={handleClick}
+      onKeyDown={handleClick}
+      role="button"
```
nigellima, line 35: Wouldnt it be better to add this to the NavBarLink component (it also has access to the navbar context) instead of making the NavBarCollapse an interactive element?

nigellima (changes requested):

src/components/Navbar/NavbarCollapse.tsx
dhavalveera: Make Sense! Thanks @nigellima . Made the changes accordingly and updated the changes with push

src/components/Navbar/NavbarLink.tsx
```diff
@@ -44,6 +48,7 @@ export const NavbarLink: FC<NavbarLinkProps> = ({
           theme.disabled[disabled ? 'on' : 'off'],
           className,
         )}
+        onClick={handleClick}
```
nigellima, line 51: Ideally I think it would be better to "merge" the user's `onClick` with your custom `handleClick` so that the user's callback does not override the behaviour you're adding

nigellima (changes requested):

src/components/Navbar/NavbarLink.tsx
dhavalveera: @nigellima would it be possible to share the example? as I've seen the other components in `/src/components/*` no where I've seen that the existing code is handling user's `onClick` event as well with our custom `handleClick` function.

nigellima: Something similar to this https://github.com/themesberg/flowbite-react/blob/main/src/components/Dropdown/DropdownItem.tsx#L53

We have an internal callback for the item click but we also make sure that the user's callback is called

dhavalveera: @nigellima  in that scenario, we've to modify the interface as well to accept the user's `onClick` as well, correct? should we do that? if so then I'll modify and push the code again.

but one question or use-case though, generally do we need an `onClick` event on `Navbar.Link`?

nigellima: it would just be something like this, with the correct types

```javascript
const handleClick = (event) => {
    setIsOpen(!isOpen);
   props.onClick && props.onClick(event)
}
```

dhavalveera: @nigellima, as we're writing it in TypeScript, so we've to add `onClick` to the interface to accept the props, as it can throw an error during build if we don't pass the `onClick` in props:

https://github.com/themesberg/flowbite-react/blob/main/src/components/Navbar/NavbarLink.tsx#L16


more specifically, in the above interface

dhavalveera:
```tsx
export interface NavbarLinkProps extends ComponentProps<'a'>, Record<string, unknown> {
  active?: boolean;
  as?: ElementType;
  disabled?: boolean;
  href?: string;
  theme?: DeepPartial<FlowbiteNavbarLinkTheme>;
  onClick: () => void
}
```

something this way need to add if we need handle user's `onClick` @nigellima

nigellima: Is it not getting it already from the `ComponentProps<'a'>`?

dhavalveera: Sorry, it's getting from `ComponentProps<'a'>`.

I've made the changes

src/components/Navbar/NavbarLink.tsx
```diff
@@ -28,14 +28,16 @@ export const NavbarLink: FC<NavbarLinkProps> = ({
   children,
   className,
   theme: customTheme = {},
+  onClick: userOnClick,
```
nigellima, line 31: No need to rename. Use just `onClick`

nigellima (changes requested):

src/components/Navbar/NavbarLink.tsx
dhavalveera: I've just renamed on differ the actual DOM's `onClick` event, @nigellima , but let me rename it to original

nigellima: Excellent. Thanks @dhavalveera

nigellima (approved):
LGTM

nigellima:
> what is LGTM @nigellima ?

_Looks Good To Me_ 😄

dhavalveera:
> > what is LGTM @nigellima ?
> 
> _Looks Good To Me_ 😄

Understood!.. can you please approve the review again, please @nigellima

nigellima (approved):

dhavalveera:
@SutuSebastian can you please review the PR, and approve?

src/components/Navbar/NavbarLink.tsx
```diff
@@ -28,12 +28,18 @@ export const NavbarLink: FC<NavbarLinkProps> = ({
...
+  onClick,
   ...props
 }) => {
-  const { theme: rootTheme } = useNavbarContext();
+  const { theme: rootTheme, isOpen, setIsOpen } = useNavbarContext();
 
   const theme = mergeDeep(rootTheme.link, customTheme);
 
+  const handleClick = (event: MouseEvent<HTMLAnchorElement>) => {
+    setIsOpen(!isOpen);
```
SutuSebastian, line 39: I'd rather be explicit to closing the navbar instead of toggle-ing it (true/false).

Suggestion:
```tsx
const handleClick = (event: MouseEvent<HTMLAnchorElement>) => {
  setIsOpen(false);
  onClick?.(event);
};
```

This also makes the `isOpen` value redundant and can be removed.

SutuSebastian (changes requested):

src/components/Navbar/NavbarLink.tsx
dhavalveera: made the changes you mentioned!..

dhavalveera (commented):
made the changes @SutuSebastian

SutuSebastian (approved):
all good now

dhavalveera:
@SutuSebastian , can you please help regarding the `CI / build (pull_request)` is getting failed now..

SutuSebastian:
> @SutuSebastian , can you please help regarding the `CI / build (pull_request)` is getting failed now..

its just codecov going nuts again...

SutuSebastian (approved):

dhavalveera:
> > @SutuSebastian , can you please help regarding the `CI / build (pull_request)` is getting failed now..
> 
> its just codecov going nuts again...

Understood!.. I don't have to worry about it, right sir @SutuSebastian ?


so I guess this PR had to be merge manually, if I am not wrong due to this check fail due to Codecov

SutuSebastian:
> > > @SutuSebastian , can you please help regarding the `CI / build (pull_request)` is getting failed now..
> > 
> > 
> > its just codecov going nuts again...
> 
> Understood!.. I don't have to worry about it, right sir @SutuSebastian ?
> 
> so I guess this PR had to be merge manually, if I am not wrong due to this check fail due to Codecov

no, nothing from ur side, i'll merge it anyway, small changes

PR #1263, opened 2024-02-06
fix(dropdown.tsx): handling the Placeholder if the label isnt provided

### LINKED ISSUE #1260: Support for placeholder in Dropdown

## Summary

Now the Dropdown has a placeholder property but it seems not being used.
Does it make sense to enable it to be used when the label is not defined?

Maybe I am using it as a select incorrectly, but I don't like the style of the Select floating menu...
If there is another way, any advise will be more than welcome.

## Context

In case a dropdown label is initially not defined, it could show the placeholder.
Now I am doing it from my components using:

<Dropdown label={label ?? placeholder} ...

dhavalveera:
currently the Dropdown - the label has to be passed, if the there is no label, then showing default placeholder

fix #1260

### DIFF
--- a/src/components/Dropdown/Dropdown.tsx
+++ b/src/components/Dropdown/Dropdown.tsx
@@ -191,7 +191,7 @@ const DropdownComponent: FC<DropdownProps> = ({
         getReferenceProps={getReferenceProps}
         renderTrigger={renderTrigger}
       >
-        {label}
+        {label ?? 'Dropdown Button'}
         {arrowIcon && <Icon className={theme.arrowIcon} />}
       </Trigger>
       {open && (


src/components/Dropdown/Dropdown.tsx
```diff
@@ -191,7 +191,7 @@ const DropdownComponent: FC<DropdownProps> = ({
         getReferenceProps={getReferenceProps}
         renderTrigger={renderTrigger}
       >
-        {label}
+        {label ?? 'Dropdown Button'}
```
SutuSebastian, line 194: this should be configurable through props

dhavalveera:
Hello @SutuSebastian , Sorry to bother you, can you please review this PR?

SutuSebastian (commented):
maybe add a new optional prop called `placeholder` then render `{label ?? placeholder}` but for this we have to make `label` optional, not sure if this is intended.

CC @rluders @tulup-conner

dawid-bobowski:
I found the same issue when using _renderTrigger_. Even though I pass a custom Component, it still requires a _label_ prop, even though it isn't used.

PR #1265, opened 2024-02-07
feat: Drawer component

dhavalveera:
* feat(drawer): created Drawer component

* This PR creates the Drawer component based on the specs of the core library

* docs(Drawer): created document page of Drawer component with examples

* docs(Drawer): improved documentation


dhavalveera:
another push of the code changes includes the position of the Drawer (top | right | bottom | left), where default one is left

rluders:
@dhavalveera when I test the example in the docs the drawer is displayed outside the example, covering the documentation left sidebar. My best guess here is that isn't desirable, right?

image

dhavalveera:
> @dhavalveera when I test the example in the docs the drawer is displayed outside the example, covering the documentation left sidebar. My best guess here is that isn't desirable, right?
> 
> image

Yes @rluders it's not desirable, but I'm not sure how avoid this, maybe change the position or something, would it be possible @rluders  Sir, please guide me? 🙏

rluders:
> Yes @rluders it's not desirable, but I'm not sure how avoid this, maybe change the position or something, would it be possible @rluders Sir, please guide me? 🙏

TBH, I have no idea. :)

dhavalveera:
@SutuSebastian , I've tried your suggestion as per shared in Discord Channel, by adding the `position: relative` to the Code Demo Wrapper Code and adding the `position: absolute` to the Drawer Code, and it worked, as well as I tried to check the Modal & Popover code in my local, and it seems working for me, can you please have a look from the above deployed Vercel Link, please 🙏

nigellima:
@dhavalveera would it be possible to have a slide transition? like, when it's left-right it would be cool to have it slide over from outside of the screen instead of just popping up instantaneously

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,82 @@
...
+  open: isOpen = false,
+  theme: customTheme = {},
+  position = 'left',
+  onClose,
+  ...props
+}) => {
+  const theme = mergeDeep(getTheme().drawer, customTheme);
+
+  useEffect(() => {
+    const handleEscapeKey = (event: KeyboardEvent) => {
```
nigellima, line 50: You would probably need to check if there is no input element within the drawer.
Because if the user has any sort of textarea/input/combobox inside the drawer, this code would close while they are typing

src/components/Drawer/DrawerHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+          <MdHome aria-hidden className={twMerge(theme.inner.titleIcon)} />
+        )}
+        {title}
+      </h5>
+
+      <button
+        type="button"
+        className={twMerge(theme.inner.closeButton)}
+        title="Close"
+        aria-label="Close Drawer"
```
nigellima, line 59: I think it's important for the users to be able to override the aria-label for this button here specifically

src/components/Drawer/Drawer.tsx
```diff
@@ -0,0 +1,82 @@
...
+        onClose?.();
+      }
+    };
+
+    document.addEventListener('keydown', handleEscapeKey);
+  }, [onClose]);
+
+  return (
+    <DrawerContext.Provider value={{ theme, onClose, isOpen }}>
+      <div
```
nigellima, line 61: Missing `role="dialog"` and `aria-describedby` for this div

src/components/Drawer/DrawerHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+}) => {
+  const id = useId();
+
+  const { isOpen, onClose, theme: rootTheme } = useDrawerContext();
+
+  const theme = mergeDeep(rootTheme.header, customTheme);
+
+  return (
+    <div className={twMerge(className)} {...props}>
+      <h5 className={twMerge(theme.inner.titleText)}>
```
nigellima, line 46: Add an `id` in this heading and set the `aria-describedby` of the main drawer div with the value of this id. You can share this information via the context you already have

nigellima (changes requested):

dhavalveera:
> @dhavalveera would it be possible to have a slide transition? like, when it's left-right it would be cool to have it slide over from outside of the screen instead of just popping up instantaneously

OK, let me check it out and see

src/components/Drawer/Drawer.tsx
dhavalveera: but in this we're checking the specific key, isn't it?

src/components/Drawer/DrawerHeader.tsx
dhavalveera: would it be possible to give a code example? as if we use `<Drawer.Header />` component this way, then how we can use users `arial-label`? suggestions please 🙏

dhavalveera:
Hello @SutuSebastian , Sorry to bother you, but can you please help me in this PR?

src/components/Drawer/Drawer.tsx
SutuSebastian: more exact: check if there is any input element focused, of any kind before executing any action on "ESC" key press.

dhavalveera: any idea how to check it @SutuSebastian ?

tulup-conner:
Thank you @dhavalveera for getting this started! I just opened a new PR branching off your changes at #1342 and added credits pointing directly to your original work here. I'm going to go ahead and close this PR since all remaining work will be conducted there.

PR #1289, opened 2024-03-08
Fix CSP violation in Carousel drag-scroll helper

rnicholus:

Users of flowbite-react with a strict CSP will not permit inline styles. In order to whitelist inline styles in a CSP, you must add the `unsafe-inline` value to the `style-src` directive. Allowing inline styles opens the application up to cross-site scripting attacks (XSS). 

The `<Carousel>` component in flowbite-react indirectly introduces inline styles via the `react-indiana-drag-scroll` library, which [inlines all of its styles via the build process](https://github.com/Norserium/react-indiana-drag-scroll/blob/6b76dcbddfcbc69a48411f91f68f266901e24694/rollup.config.js#L35).

I considered updating react-indiana-drag-scroll's build config to avoid inlining the styles, but that would introduce a signifiant breaking change for all other users of the library. So, given the small size of the dependency's codebase (1 file) and the fact that the library is already stable (no changes for 2+ years), I felt it was best to simply hoist the code into react-flowbite and do away with the inline styles by leaning on TailwindCSS classes instead. react-indiana-drag-scroll's MIT license fully permits this move.

The 2 new dependencies were dependencies of react-indiana-drag-scroll. I attempted to limit changes to that code as much as possible, to avoid regressions. But some changes were needed to address improper use of TypeScript in that project.

I did some basic manual tests in the development environment on the carousel page, and everything appears to work as expected.

I did not add tests as the dependency itself also did not have any tests to speak of.

### DIFF
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -1,3 +1,18 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+@layer utilities {
+  .hide-scrollbars::-webkit-scrollbar {
+    display: none !important;
+    height: 0 !important;
+    width: 0 !important;
+    background: transparent !important;
+    -webkit-appearance: none !important;
+  }
+  .hide-scrollbars {
+    overflow: hidden;
+    overflow: -moz-scrollbars-none;
+    scrollbar-width: none;
+  }
+}
--- a/styles/globals.css
+++ b/styles/globals.css
@@ -1,18 +1,3 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
-
-@layer utilities {
-  .hide-scrollbars::-webkit-scrollbar {
-    display: none !important;
-    height: 0 !important;
-    width: 0 !important;
-    background: transparent !important;
-    -webkit-appearance: none !important;
-  }
-  .hide-scrollbars {
-    overflow: hidden;
-    overflow: -moz-scrollbars-none;
-    scrollbar-width: none;
-  }
-}
--- a/package.json
+++ b/package.json
@@ -55,10 +55,11 @@
   },
   "dependencies": {
     "@floating-ui/react": "^0.26.2",
+    "classnames": "^2.5.1",
     "contentlayer": "^0.3.4",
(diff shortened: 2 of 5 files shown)


SutuSebastian:
Thanks for the digging! 

I was thinking more like deprecating altogether `react-indiana-drag-scroll`, and start using something more new, stable and maintained.

@rluders @tulup-conner what do u guys think?

rluders:
This is a hard question... I have no strong opinion about it. AFAIR, using the `react-indiana-drag-scroll` was 'cause it was very stable and pretty direct to solve the need that we had. I'm not against accepting this PR, nor searching a new library to replace the current dependency. I think that I would like to have some options to consider, some kind of cons and pros are always helpful in deciding what we want to do.

Also, let me point out that, IMHO, I would not care that much about introducing breaking changes before we achieve the `1.0.0` release. I guess that we still at some point can experiment and do these kinds of things, however, I do understand that the pre-release version is aiming to leave the current library and API as stable as possible until we arrive at the official release.

@SutuSebastian do you have any library in mind? What would be the benefits if compared with the proposal on this PR?

SutuSebastian:
For any swipe/scroll gestures, I'd definitely use either https://swiperjs.com/ or https://dndkit.com/.

SutuSebastian:
Or why not, since this is about carousel, https://www.embla-carousel.com/.

rnicholus:
Hey everyone, thanks for the quick responses.

The change I have here fixes an immediate issue for users of your library, and merging this doesn’t prevent you from coming back later and replacing the carousel component or the drag-scroll logic if you see a need. 

Just my 2 cents.

styles/globals.css
```diff
@@ -1,3 +1,18 @@
 @tailwind base;
 @tailwind components;
 @tailwind utilities;
+
+@layer utilities {
```
SutuSebastian, line 5: does this mean the users have to implement the following code as well?

rnicholus: good catch, fixed in 4386d62e5b55a14baa670cab09175d64ef78f318

Issue #962, opened 2023-09-14
Button can't be rendered as it's return type is ReactNode.


DavideFrancescon:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Steps to reproduce

1. Create a CRA, 
2. Install flowbite react
3. Add a button to a page

## Current behavior

```
TS2786: 'Button' cannot be used as a JSX component.
  Its return type 'ReactNode' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'.
```

## Expected behavior

To normally render a Button

## Workaround

CTRL/CMD+click on Button component will bring you to the .d file, there change the last NodeElement into JSX.Element. After this restart the server

I might fix this bug myself when i have 10 free minutes if noone picks this up before me

PadowYT2:
What React version are you using?

DavideFrancescon:
> What React version are you using?

"react": "^18.2.0"

I think the issue is caused by this, but the fix is really fast, just need to change the component return type to JSX.Element

PadowYT2:
> > What React version are you using?
> 
> "react": "^18.2.0"
> 
> I think the issue is caused by this, but the fix is really fast, just need to change the component return type to JSX.Element

Well if you can change it within your project and it works, go for it, but what about @types/react?

congon4tor:
I'm facing this issue too.

> 'Button' cannot be used as a JSX component.
  Its return type 'ReactNode' is not a valid JSX element.
  
This started happening when I updated to version 0.6.0

mrjohnr:
Hi
I've created today a typescript+react template 
` npx create-react-app my-app --template typescript`

then followed steps from there  [https://www.flowbite-react.com/docs/getting-started/introduction](url) I've installed latest version of  flowbite-react (0.7.3)

I've put a button on my app 


```
import { Button } from 'flowbite-react';

export default function MyPage() {
  return (
    <div>
      <Button>Click me</Button>
    </div>
  );
}
```
and got 

```
TS2786: 'Button' cannot be used as a JSX component.
  Its return type 'ReactNode' is not a valid JSX element.
    Type 'undefined' is not assignable to type 'Element | null'.

```

SutuSebastian:
Looking into it, will push and release a fix ASAP.

rnicholus:
Note that this also affects `Dropdown.Item`. Regression from #1244.

[...]

PR #1244, opened 2024-01-24
fix(button): Fixed Typescript error when using Next.js Link in the button `as` prop

nigellima:
- [x] I have followed the [Your First Code Contribution section of the Contributing guide](https://github.com/themesberg/flowbite-react/blob/main/CONTRIBUTING.md#your-first-code-contribution)

Summarize the changes made and the motivation behind them.
This PR fixes the Typescript error when the Next `Link` component was passed to the Button `as` prop.
The error is now fixed and the props inference of the component/element passed to the `as` prop still works properly

Reference related issues using `#` followed by the issue number.

fix #1002 fix #1107

Issue #1107, opened 2023-11-05
Button as a link error


mnakhli:
When I use the button as a link, I get an error:
<img width="1583" alt="Screenshot 1402-08-14 at 16 13 55" src="https://github.com/themesberg/flowbite-react/assets/21153535/bb69053f-ca71-440f-9424-3c35483b0a3c">

I proceeded exactly according to [the documentation](https://www.flowbite-react.com/docs/components/button#override-button-base-component).

jonnyruffles:
@rluders I have the same issue using Webstorm as my IDE. Here's some information:

![image](https://github.com/themesberg/flowbite-react/assets/136707592/eb9e3a7d-e1fc-453c-8f35-dcdf95c150a5)

```
  "dependencies": {
    < ... >
    "flowbite": "^2.2.0",
    "flowbite-react": "0.7.2",
  }
```

```
  "devDependencies": {
    "@types/node": "^20",
    "@types/react": "^18",
    "@types/react-dom": "^18",
    "autoprefixer": "^10.0.1",
    "eslint": "^8",
    "eslint-config-next": "14.0.4",
    "eslint-config-prettier": "^9.1.0",
    "postcss": "^8",
    "prettier": "^3.1.1",
    "prisma": "^5.7.0",
    "tailwindcss": "^3.3.0",
    "ts-node": "^10.9.2",
    "typescript": "^5"
  }
```

jonnyruffles:
Duplicate of #1002

Issue #1002, opened 2023-09-26
Button component `as={Link}` triggers typecheck error


shixiao:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Steps to reproduce

1. Go to ['this sandbox'](https://codesandbox.io/p/sandbox/button-as-3jpqss?file=%2Fpages%2Findex.tsx%3A9%2C63)
2. Run task `dev` (which does `yarn dev`)

## Current behavior

The second button (`<Button as={Link} href="www.example.com">Toggle 2</Button>`) triggers a typechecker error:
```
Type '{ children: string; as: ForwardRefExoticComponent<Omit<AnchorHTMLAttributes<HTMLAnchorElement>, keyof InternalLinkProps> & InternalLinkProps & { ...; } & RefAttributes<...>>; href: string; }' is not assignable to type 'Pick<Omit<AnchorHTMLAttributes<HTMLAnchorElement>, keyof InternalLinkProps> & InternalLinkProps & { ...; } & RefAttributes<...>, "slot" | ... 263 more ... | keyof InternalLinkProps>'.
  Types of property 'as' are incompatible.
    Type 'ForwardRefExoticComponent<Omit<AnchorHTMLAttributes<HTMLAnchorElement>, keyof InternalLinkProps> & InternalLinkProps & { ...; } & RefAttributes<...>>' is not assignable to type 'Url | undefined'.typescript(2322)
```

## Expected behavior

No typechecker error.

## Context

https://github.com/themesberg/flowbite-react/pull/885 made a series of Button and ButtonProps types / interfaces generic, attempting to allow `as={...}`. For some reason, this is not working for `as={Link}`.

shixiao:
https://codesandbox.io/p/sandbox/dropdownas-f992mv?file=%2Fpages%2Findex.tsx%3A6%2C29

`as={Link}` and `ref={ref}` are also not working on `Dropdown.Item`. (forwardRef used to work on them)

cc: @nigellima @rluders

rluders:
This impacts #1001

pangolingo:
I've avoided this in my codebase by creating a custom component that passes props to the button component:

```
import {
  Button as FlowbiteButton,
  ButtonProps as FlowbiteButtonProps,
} from "flowbite-react";
import Link from "next/link";

export default function Button(
[...]

PR #1298, opened 2024-03-14
fix: Revert #1244 to address regression w/ Button and DropdownItem

rnicholus:
In #1244, the return type of `Button` and `DropdownItem` was changed to `ReactNode`. This is not an acceptable return type for functional components. The correct return type is `JSX.Element`. This change fixes the regression from that PR (introduced in v0.7.3) by reverting that PR and accounting for related changes that followed that PR. Also fixes https://github.com/themesberg/flowbite-react/issues/962.

Perhaps there is a way to fulfill the goals of #1244 and solve this regression, but I wasn't able to find time to address that, and #1244 seems flawed enough that perhaps it just needs to be redesigned.


rluders:
@coderabbitai review

rnicholus:
Uh, whatever AI you’ve unleashed on this PR is pretty low-quality. All I’ve done here is ‘git revert’ 1 set of changes and partially revert another.

SutuSebastian:
> Uh, whatever AI you’ve unleashed on this PR is pretty low-quality. All I’ve done here is ‘git revert’ 1 set of changes and partially revert another.

Yep, I know that, but seems like the AI does not

rluders:
@rnicholus AI is just an tool, we still reviewing as humans. 😁

rnicholus:
Is there anything else i can do to ensure this change is merged & released?

components/homepage/components-section.tsx
```diff
@@ -25,6 +25,7 @@ export const ComponentsSection: FC = () => {
           ))}
         </div>
         <div className="mb-4 flex w-full justify-center text-center">
+          {/* @ts-expect-error TODO: fix `as` inference */}
```
nigellima, line 28: As you are fixing #962, is it necessary to have these @ts-expect-error comments?

src/components/Dropdown/Dropdown.spec.tsx
```diff
@@ -167,7 +167,7 @@ describe('Components / Dropdown', () => {
   });
 
   describe('Dropdown item render', async () => {
-    it('should override Dropdown.Item base component when using `as` prop', async () => {
+    it('should override Dropdownn.Item base component when using `as` prop', async () => {
```
nigellima, line 170: Keep the latest changes here

nigellima (changes requested):

components/homepage/components-section.tsx
rnicholus: Yes, this is needed since the fix is reverting #1244.

nigellima (approved):

src/components/Dropdown/Dropdown.spec.tsx
rnicholus: fixed in https://github.com/themesberg/flowbite-react/pull/1298/commits/3292c56f764797bd21f38aa399e590ff2553de59

rluders:
@rnicholus please, no merge with main... rebase it instead!

rnicholus:
> @rnicholus please, no merge with main... rebase it instead!

this can all be squashed on merge, correct?

rluders:
@rnicholus yes, but it is preferable to rebase with the main instead of merging it. But, anyway...

I'm wondering if this PR is actually fixing the issue, and not just reverting for some broken state. Isn't better if we could address the problem with a definitive fix? What would be the pros and cons of accepting this to fix #962 and reverting the fix for #1002 and #1107?

rnicholus:
> @rnicholus yes, but it is preferable to rebase with the main instead of merging it. But, anyway...
> 
> I'm wondering if this PR is actually fixing the issue, and not just reverting for some broken state. Isn't better if we could address the problem with a definitive fix? What would be the pros and cons of accepting this to fix #962 and reverting the fix for #1002 and #1107?

I'm not sure. My motivation here was to fix the typing issue without having to resort to `ts-ignore` in my project, and i suspect a bunch of other users are in the same boat.

>Isn't better if we could address the problem with a definitive fix?

absolutely, but i'm not sure what that fix is as i don't really have the cycles to investigate

SutuSebastian:
Closed in favor of #1308

Issue #1039, opened 2023-10-03
DatePicker component defaultDate/clear not working as expected


kresimirlesic:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Steps to reproduce

Create a default DatePicker component.

## Current behavior

Component renders with today's date as the selected one, additionally, clearing out the date sets the selected date to today.

## Expected behavior

Component input should be empty. Ideally, placeholder should be displayed if passed in.

rluders:
Well, I guess that the Datepicker references that I got were always displaying the "today" date as default. So... It is kind of, not a bug, 'cause it is working as it was programmed to do... but, got it... I guess that makes sense to level it empty sometimes.

kresimirlesic:
Hey @rluders , thank you for the quick response!

I did have to switch to react-datepicker for the time being until this gets resolved.

The way they do it is with a isClearable props, once you set it to true, you can clear out (and assign) null/undefined values.

... just a food for thought

mcle0463:
so if you pass a defaultDate prop then the DatePicker should render with that date selected? Is that the expected behavior? if so i am having this issue as well

ddiasfront:
Okay, I'm having the same issue here, if defaultDate should be taken into consideration in after-mounting phases.

I'm submitting a PR related to that.

MiroslavPetrik:
One workaround for controlled Datepicker is that you can set `value` and `placeholder` props like this:

```ts
  const emptyProps = !value
    ? {
        value: "",
        placeholder,
      }
    : {};

<Datepicker {...emptyProps } />
```

The `""` value overrides the default formatted date, which will enable the native `placeholder` to be displayed...

But yeah, the Datepicker will still have its internal value.
Good idea to make the Datepicker controlled as well.

taskrider2022:
A isClear callback will be nice to set the date to null|undefined 

In my case I will add a datepicker that is not mandatory to fill

PR #1300, opened 2024-03-15
fix(Datepicker): expose focus() and clear() methods via ref.

### LINKED ISSUE #1299: Datepicker should `forwardRef`

## Summary

It should be possible to forward `ref` to the Datepicker, in order to focus its input, or clear the value.

Currently the Datepicker is uncontrolled (only accepting `defaultValue`, not `value` prop) https://github.com/themesberg/flowbite-react/issues/1039.
Alternative to controlled component is programmatic API exposing methods to the ref object via [ `useImperativeHandle`](https://react.dev/reference/react/useImperativeHandle).

### Scenario - Clear Datepicker
I have a [controlled Datepicker ](https://form-atoms.github.io/flowbite/?path=/story/datepickerfield--optional) which a problematic scenario:
1. select a value in the datepicker
2. click reset
3. the controlled value is reset, the placeholder is displayed (everything ok on the outside)
4. open the datepicker again
5. the datepicker displays the value from the step 1, expected would be to see the `defaultValue` - a `clear()` method needs to be called.

### Scenario - Focus the datepicker
My controlled form fields, have `focus()` method, where I can focus the field when its appropriate - e.g. error is detected while submitting the form.

Currently I cannot propagate my `ref` to the `Datepicker`. So the Datepicker should accept the ref, and delegate the `focus` call into the 
internal `inputRef` as `inputRef.current.focus()` called in `useImperativeHandle`.

## Additional 
Possibly other methods, like `blur()` could be added, I don't have real use case for this right now so it's optional.

MiroslavPetrik:
Closes #1299 (possibly #1039)

# Summarize the changes made and the motivation behind them.

Enhanced the Datepicker component to expose `clear()` and `focus()` ref methods.
This is required when datepicker value is in controlled mode.

# Reference related issues using `#` followed by the issue number.
Closes #1299 and (possibly #1039)

No breaking changes.

## Summary by CodeRabbit

- **New Features**
	- Enhanced the Datepicker component to support direct manipulation through `focus` and `clear` methods.
- **Refactor**
	- Improved the Datepicker component's architecture for better usability and control.
- **Tests**
	- Added new tests to ensure the functionality of focusing and clearing values in the Datepicker component.

### DIFF
--- a/src/components/Datepicker/Datepicker.tsx
+++ b/src/components/Datepicker/Datepicker.tsx
@@ -1,7 +1,7 @@
 'use client';
 
-import type { FC, ReactNode } from 'react';
-import { useEffect, useRef, useState } from 'react';
+import type { ForwardRefRenderFunction, ReactNode } from 'react';
+import { forwardRef, useEffect, useImperativeHandle, useRef, useState } from 'react';
 import { HiArrowLeft, HiArrowRight, HiCalendar } from 'react-icons/hi';
 import { twMerge } from 'tailwind-merge';
 import { mergeDeep } from '../../helpers/merge-deep';
@@ -71,6 +71,17 @@ export interface FlowbiteDatepickerPopupTheme {
   };
 }
 
+export interface DatepickerRef {
+  /**
+   * Focus the datepicker input.
+   */
+  focus: () => void;
+  /**
+   * Clears the datepicker value back to the defaultDate.
+   */
+  clear: () => void;
+}
+
 export interface DatepickerProps extends Omit<TextInputProps, 'theme'> {
   open?: boolean;
   inline?: boolean;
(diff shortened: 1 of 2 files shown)


src/components/Datepicker/Datepicker.tsx
```diff
@@ -325,4 +355,6 @@ export const Datepicker: FC<DatepickerProps> = ({
   );
 };
 
+export const Datepicker = forwardRef(DatepickerRender);
```
MiroslavPetrik, line 358: I'm using the forwrdRef call separatelly, as having it together with the DatepickerRender adds unnecessary indentation.

SutuSebastian: agreed

anshul-kai:
Calling `clear()` doesn't clear the rendered value. For folks looking for an interim hack, try the following approach using a `key`.

```
  const ref = useRef<DatepickerRef>(null);
  const [key, setKey] = useState(Date.now());

  useEffect(() => {
    if (value === null) {
      ref.current?.clear();
      setKey(Date.now());
    }
  }, [value]);

  return (
    <Flowbite theme={{ theme: customTheme }}>
      <Datepicker
        ref={ref}
        key={key}
        value={value}
        onChange={onChange}
      />
    </Flowbite>
  );
```

Issue #1366, opened 2024-04-16
Datepicker is not allowing clearing date


bvedad:
- [X] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [X] I have tested the latest version

## Summary

Currently, the Datepicker component in Flowbite-React allows users to select a date, but lacks a fully functional clear button that can reset the date to null. The clear button only resets the date to the original value. I propose adding a flag like isClearAction to the onSelectedDateChanged callback function. This flag would be set to true when the clear button is used, allowing developers to explicitly handle the clear action by setting the date to null if desired.

Example implementation:
```
<Controller
  control={control}
  name="birthday"
  render={({ field: { onChange, value } }) => (
    <Datepicker
      onSelectedDateChanged={(date, isClear) => {
        if (isClear) {
          onChange(null); // Explicitly handle clear action
          return;
        }
        onChange(date); // Handle date change normally
      }}
      placeholder="Please enter date"
      ref={elementRef}
      value={value ? moment(value).format('YYYY-MM-DD') : ''}
    />
  )}
/>

```
This feature would enable developers to easily implement scenarios where a user might need to completely remove a previously selected date.

## Context

In various applications, especially in forms involving date entries like event planning, user registrations, or bookings, having the capability to clear a date completely is crucial. Currently, without this feature, users of our application cannot remove a date once selected; they can only change it. This limitation affects the flexibility and user experience of the application. By implementing this feature, we will provide a more robust and intuitive interaction for users who need to reset their date selections entirely.

OrlandisCS:
That's right, it's a problem that I faced and I had to change the date selector for something so basic, I hope it can be solved since I really like this design

MAKINWARA:
Could you please provide any updates on this?

lyczkop:
I had same issue and for me works when I added `key` props to `Datepicker`. That forced change in UI.
```
<Datepicker
   key={value?.toISOString()}
.....
/>
```

PR #1367, opened 2024-04-16
…

bvedad:

Update Datepicker onSelectedDateChanged signature to include isClearAction parameter

#1366

### DIFF
--- a/packages/ui/src/components/Datepicker/Datepicker.tsx
+++ b/packages/ui/src/components/Datepicker/Datepicker.tsx
@@ -96,7 +96,7 @@ export interface DatepickerProps extends Omit<TextInputProps, "theme"> {
   language?: string;
   weekStart?: WeekStart;
   theme?: DeepPartial<FlowbiteDatepickerTheme>;
-  onSelectedDateChanged?: (date: Date) => void;
+  onSelectedDateChanged?: (date: Date, isClearAction: boolean) => void;
 }
 
 const DatepickerRender: ForwardRefRenderFunction<DatepickerRef, DatepickerProps> = (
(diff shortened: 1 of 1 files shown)


rluders (approved):

rluders:
> `onChange` prop should be returning `Date | null` in the first place, avoiding this workaround

I'm not sure... 🤔  Let me think.

SutuSebastian:
Also, this has `packages/ui` changes, which should introduce a changeset, please run `bun changeset` in the root folder and follow instructions to bump the `flowbite-react` package to a `patch`, in the end u have to give it a changelog/summary of the PR, then commit the generated changeset.

rluders (changes requested):
Check comments ☝️

bvedad:
> `onChange` prop should be returning `Date | null` in the first place, avoiding this workaround

@SutuSebastian I agree with you but the reason why I proposed this implementation is because at the moment clear button does not clear the date to be null, but it resets it to the default value. To me this is counterintuitive but if developers decide for some reason to implement it that way I am not planning to be a bottleneck. This approach is harmless and it allows us to set dates to null if we want. If you want to I can implement it so onSelectedDateChanged accept Date | null but then I will need to change implementation of clear from:
```
const clearDate = () => {
    changeSelectedDate(defaultDate, true, true);
    if (defaultDate) {
      setViewDate(defaultDate);
    }
  };
```
to something like:
```
const clearDate = () => {
    changeSelectedDate(defaultDate, true, true);
    setViewDate(defaultDate);
  };
```

and as well change code from:
```
{showClearButton && (
  <button
    type="button"
    className={twMerge(theme.popup.footer.button.base, theme.popup.footer.button.clear)}
    onClick={() => {
      changeSelectedDate(defaultDate, true, true);
      if (defaultDate) {
        setViewDate(defaultDate);
      }
    }}
  >
    {labelClearButton}
  </button>
)}
```
to something like below:
```
{showClearButton && (
  <button
    type="button"
    className={twMerge(theme.popup.footer.button.base, theme.popup.footer.button.clear)}
    onClick={() => {
      changeSelectedDate(null, true, true);
      setViewDate(null);
    }}
  >
    {labelClearButton}
  </button>
)}
```

bvedad:
> Also, this has `packages/ui` changes, which should introduce a changeset, please run `bun changeset` in the root folder and follow instructions to bump the `flowbite-react` package to a `patch`, in the end u have to give it a changelog/summary of the PR, then commit the generated changeset.

Thanks for giving me this information. I have just generated changelog content and pushed it to this MR.

bvedad:
> Check comments ☝️

I have just checked them and I have performed the necessary changes.

Issue #1257, opened 2024-02-02
"Open" prop for Tooltip component


cthacker-udel:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Summary

[Describe how it should work, and provide examples of the solution, which might include screenshots or code snippets.](https://floating-ui.com/docs/useFloating#open)

## Context

I am trying to make the tooltip only appear when form errors are apparent. There are no options available in flowbite/react, so I am trying to use base flowbite, https://flowbite.com/docs/components/tooltips/#options, to no avail. Tooltips are not extremely necessary for my application, but would be incredibly useful to not have to scrap extra functionality because of one prop missing.

PR #1383, opened 2024-04-30
feat: make tooltips stay open

lirbre:
Hey, 

I'm pretending to be more active again giving contributions to flow-bite. It has some time from my last contributions so I'm feeling a little ankward about the standards of this project lol

Just picked up #1257 and gave it a shot. To summarize, I'm adding the possibility to override the open/closed status for tooltips: 

image

I implemented it in a simple way, just added a new prop called "isOpen" that overrides the "open" state - now called innerOpen. Since he also wanted to change it's trigger, I made it's default to "null" and did a simple check to ensure that it would work as expected when true or false. 

Tbh idk if it's a valid way to do it here at flow-bite so let's discuss about it. :)

Cheers,


SutuSebastian:
it seems like ur branch base HEAD is way out of date, leading to massive conflicts.

Please sync ur fork with latest `themesberg:main` and then rework/reopen this PR, its easier this way.

brenoliradev:
Yeah. I just noticed it yesterday.

I was using my outdated main (on my personal repo) instead of upstream main. 

Just fixed it!

packages/ui/src/components/Floating/Floating.tsx
```diff
@@ -59,10 +60,17 @@ export const Floating: FC<FloatingProps> = ({
...
   trigger = "hover",
   minWidth,
+  isOpen = null,
   ...props
 }) => {
   const arrowRef = useRef<HTMLDivElement>(null);
-  const [open, setOpen] = useState(false);
+  const [innerOpen, setOpen] = useState(false);
+
+  const open = useMemo(() => {
```
SutuSebastian, line 69: no need to use `useMemo()` here since its not a heavy calculation or any iteration, we can just stick to a more simplified version:

```tsx
  const [_open, setOpen] = useState(false);

  const open = isOpen ?? _open;
```

Issue #1352, opened 2024-04-05
Theme sync may cause render loops across different browser tabs


lancegliser:
- [X] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [X] I have tested the latest version

## Steps to reproduce

Clone [reproduction repo](https://github.com/lancegliser/flowbite-react-theme-mode-sync-issue). _Basically npm create flowbite-react@latest_.

1. Run dev server
2. Open 2-3 copies in different tabs
3. Adjust a copy of one window's performance tab to use 6x throttling
4. Spam the light / dark switcher

## Current behavior

The tabs are reading, rerendering, and causing other events to communicate back and forth probably based on local storage hooks. This _can_ cause them to hit infinite loops in all tabs.

![ezgif-5-ce03d57c56](https://github.com/themesberg/flowbite-react/assets/12085479/c711da02-9696-48d7-8d08-9d331ee4f384)

## Expected behavior

No infinite loops.

## Context

The reproduction repo here is very silly. We found this bug in our real application, which renders a number of complex SVG's and other expensive operations. A user had just two tabs open, and was able to cause it with a simple double click on the theme switcher.

SutuSebastian:
> @SutuSebastian Is there any particular direction the Flowbite team might prefer? If I find hours, I could start a PR for discussion.

Feel free to contribute with whatever u can 🔥

lancegliser:
No real code yet, but just reviewing what it currently does I have some questions:

- Why is there both a local storage watcher _and_ an event watcher?
- I considered using a `debounce` around `handleSetMode`, but in automated scenario, it's possible this could break tests. They could test the page, swap dark, test the page, and swap again within the debounce timeout.

SutuSebastian:
Fixed by #1498  in [flowbite-react@0.11.0](https://github.com/themesberg/flowbite-react/releases/tag/flowbite-react%400.11.0) release.

PR #1400, opened 2024-05-23
Updated theme mode 'auto' to remove storage integration

lancegliser:

- Updated the set mode functionality to ignore and remove storage items when set to auto.
- Added a debounce to set mode (which we should debate as it might break automated tests)

#1352


packages/ui/src/hooks/use-theme-mode.ts
```diff
@@ -89,9 +91,17 @@ const useSyncMode = (onChange: (mode: ThemeMode) => void) => {
 };
 
 /**
- * Sets the give value in local storage
+ * Sets the give value in local storage.
+ * "auto" will remove the local storage value if present.
  */
-const setModeInLS = (mode: ThemeMode) => localStorage.setItem(LS_THEME_MODE, mode);
+const setModeInLS = (mode: ThemeMode) => {
+  if (mode === "auto") {
```
SutuSebastian, line 98: this will break the current logic, `auto` value is meant to be saved in LS and from that computed to either `light` or `dark`

Issue #1402, opened 2024-05-24
Dropdown.Item - MouseEvent<HTMLElement> is not passed through onclick


rlongre:
Because the MouseEvent is not passed through dropdown items, scenarios such as stopping propagation when the dropdown is descendant of another clickable item cannot be supported.

https://github.com/themesberg/flowbite-react/blob/b2640bcc70822e22b65a76e33b651ee291e12ec1/packages/ui/src/components/Dropdown/DropdownItem.tsx#L52-L55

By contrast, NavbarLink passes the event through:

https://github.com/themesberg/flowbite-react/blob/b2640bcc70822e22b65a76e33b651ee291e12ec1/packages/ui/src/components/Navbar/NavbarLink.tsx#L38-L41

PR #1474, opened 2024-08-25
Fix/dropdown label optional

khanSoliheen:
This commit resolves issue #1402 where the MouseEvent was not being passed through the DropdownItem's onClick handler. 

- Updated the DropdownItem component to correctly pass the MouseEvent to the onClick handler.
- This change allows for proper event handling, including scenarios where event propagation needs to be stopped, such as when the dropdown is a descendant of another clickable item.
- Ensured consistency with NavbarLink, which already passes the event through.

By addressing this, we improve the flexibility and usability of the Dropdown component within the UI, enabling better integration with other interactive elements.


SutuSebastian (changes requested):
LGTM! 

Please add a changelog by running `bun changelog` and mark it as a patch.

rluders:
@khanSoliheen please, check the code format, and rebase it before execute the changelog. But, aside from that, lgtm.

Issue #1469, opened 2024-08-16
fix(Carousel): can't give dynamic children to Carousel


kage1020:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Steps to reproduce

Create dynamic children Carousel component like below:

```tsx
function App() {
  const [count, setCount] = useState(0);
  return (
    <div className="bg-gray-800 w-screen h-screen">
      <Button onClick={() => setCount((p) => p + 1)}>count: {count}</Button>
      <Carousel>
        <img src={viteLogo} className="logo" alt="Vite logo" />
        <img src={reactLogo} className="logo" alt="React logo" />
        {count < 5 && <img src={viteLogo} className="logo" alt="Vite logo" />}
      </Carousel>
    </div>
  );
}
```

Then, count up to 5 by clicking button

## Current behavior

Flowbite throws error of `TypeError: Cannot read properties of null (reading 'props')` because it try to get `child.props.className` in [this code](https://github.com/themesberg/flowbite-react/blob/main/packages/ui/src/components/Carousel/Carousel.tsx#L93).

## Expected behavior

Skip processing Falsy values.

## Context

Live Preview: https://stackblitz.com/edit/vitejs-vite-fynok7?file=src%2FApp.tsx

PR #1475, opened 2024-08-26
fix: Insertion of Dynamic Child in Carousel crashes the Page/App

dhavalveera:
The updated code improves the robustness of handling dynamic children in the `Carousel` component. 

**Differences:**
1. **Type Handling**: The updated code explicitly checks if `child` is a valid React element using `isValidElement()`. This ensures type safety and prevents errors when children are not valid React elements.
2. **Conditional Handling**: The updated code now handles cases where `child` might not be a valid React element by returning it as-is or handling it differently, which prevents app crashes when dynamic content is inserted.

**Fix Summary:**
The fix addresses issues with dynamic children in the `Carousel` by ensuring only valid React elements are processed with `cloneElement`, thereby avoiding crashes and improving stability.

`Old Code Error:`

image

`Fixed Image`:

image

Now after the state value is equal or above 5, then it's not crashing anymore.

This PR points to Fix the Issue #1469


packages/ui/src/components/Carousel/Carousel.tsx
```diff
@@ -88,11 +88,15 @@ export const Carousel: FC<CarouselProps> = ({
...
-        }),
-      ),
+      Children.map(children as ReactElement[], (child: ReactElement<CarouselProps>) => {
+        if (isValidElement(child)) {
+          return cloneElement(child as ReactElement, {
+            className: twMerge(theme.item.base, child.props.className || ""),
+          });
+        }
+        // If child is not a valid React element, return it as-is or handle it accordingly
+        return child;
```
rluders, line 98: Don't we also want to include the `twMerge` for the `child.className` here? :thinking:

dhavalveera: but here, on this specific line there won't be any `child.className` because of the dynamic child, `child.props` & `child.props.className` is throwing Error @rluders

dhavalveera: @rluders & @SutuSebastian --- your inputs please

SutuSebastian:
this is why `Children.map` is not a good idea to inject props, better create primitives and export them as wrappers such as `<CarouselItem>` and use react context as much as possible.

In the next major version this component will be wiped off and started from scratch using other library.

SutuSebastian:
> > this is why `Children.map` is not a good idea to inject props, better create primitives and export them as wrappers such as `<CarouselItem>` and use react context as much as possible.
> > In the next major version this component will be wiped off and started from scratch using other library.
> 
> should I close this PR or what? @SutuSebastian

U can keep it, it will take time until we get the new version in

Issue #1489, opened 2024-09-16
Export theme.ts / create-theme directly


theogravity:
Development under Next.js slows down in dev mode when importing from the root `flowbite-react` because it has to compile all exports since we're importing directly from `flowbite-react`. 

It would really help if `theme.ts` can be directly imported:

https://github.com/themesberg/flowbite-react/blob/main/packages/ui/src/theme.ts

And same with the helpers / create-theme:

https://github.com/themesberg/flowbite-react/blob/main/packages/ui/src/helpers/create-theme.ts

theogravity:
I also see a reference to `react-icons` in the bundler that also may contribute to compile times too, so just adding support for more direct imports may not fix the issue entirely.

rluders:
Probably related to
#1197 #1447

theogravity:
This report is with `[optimizePackageImports: ["flowbite-react"]](https://vercel.com/blog/how-we-optimized-package-imports-in-next-js#new-solution-optimizepackageimports)` being used.

PR #1491, opened 2024-09-16
feat: add theme exports to components

theogravity:

See: https://github.com/themesberg/flowbite-react/issues/1489

Summary: Next.js has extremely long compile times when starting up in dev mode as it will compile all exports for referenced imports. A majority of us will import components directly from `flowbite-react`, meaning that Next.js will compile all the exports found in the entrypoint file. The "proper" way to import in Next.js is to import items individually via their direct path. 

```typescript
import { Table } from "flowbite-react"
```

should be instead:

```typescript
import { Table } from "flowbite-react/components/Table
```

This PR adds component-specific theme exports into the component entrypoint files.

### DIFF
--- a/.changeset/healthy-crabs-melt.md
+++ b/.changeset/healthy-crabs-melt.md
@@ -1,0 +1,5 @@
+---
+"flowbite-react": patch
+---
+
+Add theme exports to all components.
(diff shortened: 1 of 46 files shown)


SutuSebastian:
This is nice but will not work due to theme merging on a higher level applying the custom `theme` from component props. Basically to do the `theme` customisation we need people to import from the higher level one (for now), until the next big refactor.

SutuSebastian:
Fixed by #1498 .

PR #1520, opened 2024-12-26
Feat/render timeline point inner content

adel-gu:

### Summary
This PR adds an **optional** `render` prop on the `TimelinePoint` to render any component as inner content for the `TimelinePoint` component.

### Example usage
```javascript
    ...
      <Timeline.Point
          render={() => (
            <Avatar img="https://flowbite.com/docs/images/people/profile-picture-3.jpg" rounded size="xm" />
          )}
      />
    ...
```

### DIFF
--- a/packages/ui/src/components/Timeline/TimelinePoint.tsx
+++ b/packages/ui/src/components/Timeline/TimelinePoint.tsx
@@ -26,6 +26,7 @@ export interface FlowbiteTimelinePointTheme {
 export interface TimelnePointProps extends ComponentProps<"div"> {
   icon?: FC<ComponentProps<"svg">>;
   theme?: DeepPartial<FlowbiteTimelinePointTheme>;
+  render?: () => React.ReactElement;
 }
 
 export const TimelinePoint: FC<TimelnePointProps> = ({
@@ -33,6 +34,7 @@ export const TimelinePoint: FC<TimelnePointProps> = ({
   className,
   icon: Icon,
   theme: customTheme = {},
+  render = undefined,
   ...props
 }) => {
   const { horizontal } = useTimelineContext();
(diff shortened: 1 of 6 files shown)


packages/ui/src/components/Timeline/TimelinePoint.tsx
```diff
@@ -26,13 +26,15 @@ export interface FlowbiteTimelinePointTheme {
...
   theme?: DeepPartial<FlowbiteTimelinePointTheme>;
+  render?: () => React.ReactElement;
 }
 
 export const TimelinePoint: FC<TimelnePointProps> = ({
   children,
   className,
   icon: Icon,
   theme: customTheme = {},
+  render = undefined,
```
SutuSebastian, line 37: the default to `undefined` is redundant in this case

SutuSebastian (commented):
make sure to run `bun changeset` and select `flowbite-react` for a `patch` change, give it a title and add the file to this PR

packages/ui/src/components/Timeline/TimelinePoint.tsx
adel-gu: 🙈

packages/ui/src/components/Timeline/TimelinePoint.tsx
```diff
@@ -26,13 +26,15 @@ export interface FlowbiteTimelinePointTheme {
 export interface TimelnePointProps extends ComponentProps<"div"> {
   icon?: FC<ComponentProps<"svg">>;
   theme?: DeepPartial<FlowbiteTimelinePointTheme>;
+  render?: () => React.ReactElement;
```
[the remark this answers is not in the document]
adel-gu: Thanks for the reviews @SutuSebastian 👏🏼

I get the part where we need the `render` property to be more flexible. ✅

I am not clear on why we'd want to include props in the render function
>....or in case we pass some state props `(props) => React.ReactNode.

Do you mean having the `props` defined in the `render` function? Could you please elaborate with examples of how having props could be useful 🙏🏼

Issue #1521, opened 2024-12-29
Breadcrumb and Next.js's Link


phbernard:
- [x] I have searched the [Issues](https://github.com/themesberg/flowbite-react/issues) to see if this bug has already been reported
- [x] I have tested the latest version

## Summary

When used with Next.js, some components have a `as` attribute so they can be used with Next.js's `Link` component. For example:

```
import Link from 'next/link';
import { NavbarBrand } from 'flowbite-react';

<NavbarBrand as={Link} href='/'>...
```

However, it doesn't seem to be the case of `BreadcrumbItem`. It takes only an `href` element, but not `as`.

It is possible to surround a `BreadcrumbItem` with a `Link`. But it breaks the styling, a sign that this is not the right way to handle this.

Am I missing something?

SutuSebastian:
> Any news about this feedback? When posting it, my feeling was that it could be addressed quite easily. The PR is rather simple and in line with other components of flowbite-react.

Sorry, hectic days here, left a small comment on the PR btw.

insa-gra:
Hi, since I am building my application with NextJS, I would also be very interested in this feature. Is there a chance that this will still be implemented, or is there another workaround for using NextJS links as breadcrumb items?

PR #1526, opened 2025-01-25
Add 'as' prop support to BreadcrumbItem

phbernard:

As described in #1521, this PR adds a `as` prop to `BreadcrumbItem`. This prop is similar to the `as` prop in `NavbarBrand`. It is useful in Next.js, where we need to use the custom `Link` component instead of the regular `a`.

### DIFF
--- a/packages/ui/src/components/Breadcrumb/BreadcrumbItem.tsx
+++ b/packages/ui/src/components/Breadcrumb/BreadcrumbItem.tsx
@@ -1,4 +1,4 @@
-import type { ComponentProps, FC } from "react";
+import type { ComponentProps, ElementType, FC } from "react";
 import { forwardRef } from "react";
 import { HiOutlineChevronRight } from "react-icons/hi";
 import { twMerge } from "tailwind-merge";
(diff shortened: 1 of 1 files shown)


packages/ui/src/components/Breadcrumb/BreadcrumbItem.tsx
```diff
@@ -15,15 +15,16 @@ export interface FlowbiteBreadcrumbItemTheme {
...
   icon?: FC<ComponentProps<"svg">>;
   theme?: DeepPartial<FlowbiteBreadcrumbItemTheme>;
 }
 
 export const BreadcrumbItem = forwardRef<HTMLAnchorElement | HTMLSpanElement, BreadcrumbItemProps>(
-  ({ children, className, href, icon: Icon, theme: customTheme = {}, ...props }, ref) => {
+  ({ as = "a", children, className, href, icon: Icon, theme: customTheme = {}, ...props }, ref) => {
     const isLink = typeof href !== "undefined";
-    const Component = isLink ? "a" : "span";
+    const Component = isLink ? as : "span";
```
SutuSebastian, line 27: with this implementation the `as` prop will only work if the `isLink` is truthy which also infers from `href` being defined.

not quite separation of concerns here

PR #1574, opened 2025-05-21
1531: Fix/DatePicker month view error: selected month not highlighted

lydiapostk:
Fixed error where the DataPicker component Month view does not highlight the month of the selected Date.

This is because the date of newDate is automatically set to 1, and hence when compared against the selectedDate, they would always be different.

Have modified one line of code to set the newDate to have the same date as selectedDate if the latter is not null.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved month selection accuracy in the date picker by aligning the selected month with the day from the currently selected date, when available.


SutuSebastian:
Looks good!

do u think u can write some tests to ensure this bug never gets in again?

PR #1635, opened 2025-12-07
refactor(tailwindcss): enhance color handling with CSS variables

SutuSebastian:
Use CSS variables for `tailwindcss@v4` instead of inline colors.

### Changes

- [x] automatically generate `tailwindcss@v4` config file
- [x] use css variables for the `v4` config

### Breaking changes

Only applicable to `tailwindcss@v4`:

before
`@plugin "flowbite-react/plugins/tailwindcss"` - this points to the legacy JS plugin (that uses inline colors)

after
`@import "flowbite-react/plugins/tailwindcss"` - this points to the CSS-based plugin (that uses CSS variables)

The breaking change is self-managed once upgrading `flowbite-react` and starting/building the app, which runs the `flowbite-react dev` or respectively `flowbite-react build` command that triggers the patch/fix/migration automatically.

## Summary by CodeRabbit

* **New Features**
  * Added Tailwind CSS v4 support using CSS variables for colors and styling.
  * Introduced automated conversion to a CSS-based Tailwind plugin during build/dev so upgrades apply migration steps automatically.

* **Breaking Changes**
  * Tailwind v4 configuration now uses an import-based CSS plugin instead of the legacy plugin directive; configuration is migrated during upgrade/build.

### DIFF
--- a/.changeset/free-turkeys-lick.md
+++ b/.changeset/free-turkeys-lick.md
@@ -17,6 +17,6 @@ before
 `@plugin "flowbite-react/plugins/tailwindcss"` - this points to the legacy JS plugin (that uses inline colors)
 
 after
-`@import "flowbite-react/plugins/tailwindcss"` - this points to the CSS based plugin (that uses CSS variables)
+`@import "flowbite-react/plugins/tailwindcss"` - this points to the CSS-based plugin (that uses CSS variables)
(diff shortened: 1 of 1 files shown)


rickross:
Thanks, Sebastian!

PR #1641, opened 2025-12-12
fix: hide modal close button when not dismissible

ebbbang:
## Summary
Fixes the modal close button to only render when the modal is dismissible. Previously, the X button in the modal header would always display, even when `dismissible={false}`, which created a confusing UX where the button appeared but had no effect.

## Changes
- Added `dismissible` prop to `ModalContext` interface
- Updated `Modal` component to pass `dismissible` through context
- Modified `ModalHeader` to conditionally render the close button based on `dismissible` prop
- Added 2 new tests to verify the close button visibility behavior
- Updated existing test to pass `dismissible` prop where needed

## Behavior
- **Before**: Close button always visible in modal header
- **After**: Close button only visible when `dismissible={true}`

## Test Plan
- [x] All existing tests pass (11/11 Modal tests)
- [x] Added test: close button should not render when modal is not dismissible
- [x] Added test: close button should render when modal is dismissible
- [x] Manually verified: modal without `dismissible` prop shows no close button
- [x] Manually verified: modal with `dismissible={true}` shows close button

## Screenshots
N/A - behavioral change only

These follow the project's conventions from the contributing guide, including the commitizen-style prefix and the Claude Code attribution footer.

## Summary by CodeRabbit

* **New Features**
  * Modal now supports a dismissible option so the close button is shown or hidden based on configuration.

* **Tests**
  * Added tests to verify the close button is present only when the modal is dismissible and that dismissal via the button behaves correctly.

* **Chores**
  * Added a changelog entry noting the close-button behavior update.

### DIFF
--- a/.changeset/social-kings-cut.md
+++ b/.changeset/social-kings-cut.md
@@ -1,0 +1,5 @@
+---
+"flowbite-react": patch
+---
+
+Ensure modal close button respects dismissible prop
--- a/packages/ui/src/components/Modal/Modal.test.tsx
+++ b/packages/ui/src/components/Modal/Modal.test.tsx
@@ -26,6 +26,34 @@ describe("Components / Modal", () => {
     expect(modal).not.toBeInTheDocument();
   });
 
+  it("should not render close button when modal is not dismissible", async () => {
+    const user = userEvent.setup();
+
+    render(<TestModal />);
+
+    await user.click(triggerButton());
+
+    const modal = dialog();
+    expect(modal).toBeInTheDocument();
+
+    const closeButton = screen.queryByLabelText("Close");
+    expect(closeButton).not.toBeInTheDocument();
+  });
+
+  it("should render close button when modal is dismissible", async () => {
+    const user = userEvent.setup();
+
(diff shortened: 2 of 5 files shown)


SutuSebastian (changes requested):
LGTM!

One thing left:
- run `bun run changeset`
- select `flowbite-react` package
- bump as `patch`

SutuSebastian (approved):

ebbbang:
> LGTM!
> 
> One thing left:
> 
> * run `bun run changeset`
> * select `flowbite-react` package
> * bump as `patch`

Done

lyleunderwood:
>Consider the breaking change implications of the false default.
>
>With dismissible defaulting to false, existing code that doesn't explicitly pass this prop will no longer show the close button (previously it always showed). While this fixes the inconsistency described in the PR and creates more logical behavior, it could surprise users upgrading the library.

ME!!!

Took me like an hour to figure out why _all_ my close buttons disappeared, trapping users in modals! And here's the note in the changelog:

> Ensure modal close button respects dismissible prop

That's it?? In a patch release???

:sob:

PR #1649, opened 2026-01-05
Fix CLI crash when using custom config (dark: false)

aliamadi:
### Summary

The Flowbite React CLI could crash at runtime when a custom configuration
was used (e.g. `dark: false`). This triggers the CLI to evaluate the
`<ThemeInit />` warning logic, where an array method was incorrectly called
on a `MapIterator`.

### Fix

- Safely evaluate `checkedMap` values without calling array methods on a `MapIterator`
- Only run the warning logic after at least one file has been checked

```js
this.checkedMap.size > 0 &&
!Array.from(this.checkedMap.values()).some(Boolean)
```

### Verification
Built the package and tested it in a Next.js project where the CLI previously
crashed with `dark: false`. The issue no longer occurs.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a CLI crash and improved warning display so warnings appear only when appropriate.

* **Documentation**
  * Added a changelog entry and a small patch version update for a UI dependency.

### DIFF
--- a/.changeset/twelve-dogs-refuse.md
+++ b/.changeset/twelve-dogs-refuse.md
@@ -1,0 +1,5 @@
+---
+"flowbite-react": patch
+---
+
+Fix CLI crash by safely iterating over Map values in init logger
--- a/packages/ui/src/cli/utils/create-init-logger.ts
+++ b/packages/ui/src/cli/utils/create-init-logger.ts
@@ -20,8 +20,16 @@ export function createInitLogger(config: Config) {
       );
     },
     get showWarning() {
-      return this.checkedMap.size > 0 &&
-         !Array.from(this.checkedMap.values()).some(Boolean);
+      let hasChecked = false;
+
+      for (const value of this.checkedMap.values()) {
+        if (value) {
+          hasChecked = true;
+          break;
+        }
+      }
+
+      return !hasChecked;
     },
     /**
      * Checks if `<ThemeInit />` component is used in the given file content


SutuSebastian:
Please provide full reproduction steps, because the "could crash" is very vague and non technical, also I never experienced this locally or in any React based framework or technology out of all specified in the "Integrations".

It either crashes or not under certain situations, which I would like to test myself before working on this.

aliamadi:
> Please provide full reproduction steps, because the "could crash" is very vague and non technical, also I never experienced this locally or in any React based framework or technology out of all specified in the "Integrations".
> 
> It either crashes or not under certain situations, which I would like to test myself before working on this.

Thanks for checking this. Here are the exact reproduction steps:

Environment:
- OS: WSL (Ubuntu-24.04)
- Node: v20.19.5
- Package manager: pnpm

Steps:
1. Run: npx create-flowbite-react@latest -t nextjs
2. Enter the generated project directory
3. Run: pnpm install
4. Edit .flowbite-react/config.json and set:
   {
     "dark": false
   }
5. Run: pnpm dev (or pnpm build)

Result:
- The app crashes with the following error:

  TypeError: this.checkedMap.values(...).find is not a function
  at get showWarning [as showWarning]
  (flowbite-react/dist/cli/utils/create-init-logger.cjs)

Expected result:
- The app should run correctly, same as when "dark" is set to true.

Additional notes:
- This issue consistently reproduces on Node v20.19.5.
- Using newer Node versions (without changing the code) does not reproduce the crash.
- With "dark": true, everything works as expected.

I've attached a screenshot of the full error output for additional context.
Screenshot 2026-01-06 123827

SutuSebastian:
Thanks for the extra info. This is exactly what I wanted to highlight out of this, the node version has something to do with it and I need to dig into how it resolves the JS part in that exact version.

SutuSebastian:
I see it now, can confirm the issue.
Since `Bun` has built-in map iterator helpers, but can't guarantee that all the rest of the JS runtimes have it, it's safer to stick to a `for...of` loop instead, since it's much more efficient.

eg:
```tsx
get showWarning() {
  let hasChecked = false;

  for (const value of this.checkedMap.values()) {
    if (value) {
      hasChecked = true;
      break;
    }
  }

  return !hasChecked;
}
```

1. please replace ur change with the above snippet
2. make sure to create a changeset file by running `bun changeset` at the root level, and select `flowbite-react` package and bump it as a `patch`

aliamadi:
> I see it now, can confirm the issue. Since `Bun` has built-in map iterator helpers, but can't guarantee that all the rest of the JS runtimes have it, it's safer to stick to a `for...of` loop instead, since it's much more efficient.
> 
> eg:
> 
> ```tsx
> get showWarning() {
>   let hasChecked = false;
> 
>   for (const value of this.checkedMap.values()) {
>     if (value) {
>       hasChecked = true;
>       break;
>     }
>   }
> 
>   return !hasChecked;
> }
> ```
> 
> 1. please replace ur change with the above snippet
> 2. make sure to create a changeset file by running `bun changeset` at the root level, and select `flowbite-react` package and bump it as a `patch`

Thanks! I replaced the logic with the suggested for...of loop
and added a patch changeset for flowbite-react as requested.

SutuSebastian (approved):
Awesome!

PR #1660, opened 2026-03-06
fix(Modal): show header close button when onClose is provided

### LINKED ISSUE #1658: Docs wrong: onClose vs dismissible

## Steps to reproduce

Make modal according to the [official docs on Modal](https://flowbite-react.com/docs/components/modal#default-modal)

## Current behavior

There is no close button, even though onClose is specified.

## Expected behavior

The close button should either work the way it says in the docs, or the docs should be updated!

## Context

Behavior was changed in https://github.com/themesberg/flowbite-react/pull/1641 to never render the close button unless `dismissible` is set to `true`. This was clearly a _breaking_ API change that was released in `0.12.14` (a PATCH release) with only this in the changelog:

> Ensure modal close button respects dismissible prop

I accepted this insidious patch release, and all my damned close buttons disappeared, trapping my users in modals.

The PR description states this:

> Fixes the modal close button to only render when the modal is dismissible. Previously, the X button in the modal header would always display, even when dismissible={false}, which created a confusing UX where the button appeared but had no effect.

Which is clearly not true, clicking on the close button would indeed call the `onClose` handler which I was using to (😱) close modals!

I'm salty.

I'll also note that the `dismissible` property is described in the docs thusly:

> To enable the modal to be dismissed when clicking outside of the component (ie. the backdrop) then you can pass the dismissible prop to the <Modal> component from React.

That's it. Seems unrelated? My expectation, based on my reading of the docs, would be that the `onClose` prop is called when the user clicks the close button. And if one were to pass `dismissible=[true}` then clicking on the backdrop would _also_ cause `onClose` to be invoked. That's it.

KrotovM:

## Summary
The Modal header close (X) button was only shown when `dismissible={true}`, which tied the close control to backdrop/Escape behavior and broke the case where onClose is used without dismissible (see #1658).

**Changes**:
* *ModalHeader*: The close button is rendered when `onClose` is provided, regardless of `dismissible`.
* *dismissible*: Still only controls backdrop click and Escape key dismissal; behavior and docs unchanged.
* *Tests*: Updated and added cases: close button when `onClose` is provided; no close button when `onClose` is not provided; backdrop does not close when `dismissible` is true but `onClose` is not provided.
* *Docs*: Clarified in `modal.mdx` that the header close (X) button is shown when onClose is passed.

**Motivation**: Restore the expected behavior (close button when onClose is set), match the documented meaning of `dismissible`.

## Related issues
Fixes #1658.

## Breaking API changes
**None**.

No props were added or removed and the theme was not changed. This only restores and clarifies behavior:
* If you use `onClose` without `dismissible`: The header close button now appears (and works) again; no code change needed.
* If you use `dismissible={true}` and `onClose`: Behavior is unchanged; no code change needed.
* If you intentionally hid the close button by omitting `onClose`: Behavior is unchanged; no code change needed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated modal close button behavior: the header close (X) button now appears when the `onClose` prop is provided and triggers the callback when clicked.

* **Documentation**
  * Clarified modal documentation regarding close button visibility and behavior based on the `onClose` prop.

### DIFF
--- a/apps/web/content/docs/components/modal.mdx
+++ b/apps/web/content/docs/components/modal.mdx
@@ -19,7 +19,7 @@ Use the `<Modal>` React component to show a dialog element to the user with a he
 
 The visibility of the component can be set by passing a boolean value to the `show` prop on the `<Modal>` component and we recommend you to do this via the React state.
 
-Using a `openModal` and `setOpenModal` state for the main React component should allow you to easily manage the state of the Modal component and use inline functions on event listeners such as `onClick` or `onClose`.
+Using a `openModal` and `setOpenModal` state for the main React component should allow you to easily manage the state of the Modal component and use inline functions on event listeners such as `onClick` or `onClose`. When you pass the `onClose` prop, the header close (X) button is shown and calls `onClose` when clicked.
 
 <Example name="modal.root" />
 
--- a/packages/ui/src/components/Modal/Modal.test.tsx
+++ b/packages/ui/src/components/Modal/Modal.test.tsx
@@ -26,10 +26,23 @@ describe("Components / Modal", () => {
     expect(modal).not.toBeInTheDocument();
   });
 
-  it("should not render close button when modal is not dismissible", async () => {
+  it("should not close on backdrop press when dismissible but onClose is not provided", async () => {
     const user = userEvent.setup();
 
-    render(<TestModal />);
+    render(<TestModalWithoutOnClose dismissible />);
+
+    await user.click(triggerButton());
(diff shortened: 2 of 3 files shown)


rnicholus:
Is this project no longer maintained? no activity in months (that I can see)

KrotovM:
> Is this project no longer maintained? no activity in months (that I can see)

Yeah, it does look like there hasn't been much activity for a while. I'm not a maintainer, though, so I'm not sure what the current status of the project is.

PR #1690, opened 2026-09-06
fix(Pagination): show page number when totalPages is 1

### LINKED ISSUE #1048: Pagination is broken if there is only one page

## Steps to reproduce

1. Render pagination component with this props:
```
<Pagination 
  currentPage={1} 
  totalPages={1} 
  showIcons
  nextLabel=""
  previousLabel=""
/>
```

## Current behavior

There is no `1` between the arrows, just two buttons (previous and next) 
Screenshot 2023-10-06 at 11 33 51

## Expected behavior

`1` is shown between the arrows, similarly to the case when we have 2 and more pages.
image

## Context

I would like to show the `Pagination` component with only one page when user has a small amount of data.

The problem is in `range` function, where this check is contained
```
if (start >= end) {
    return [];
}
```
I would assume it should be
```
if (start > end) {
    return [];
}
```
but not sure whether this breaks the expected behaviour from your side. If you are fine I can create the PR.

Thanks,
Sergii

Abdellox:
Fixes #1048

The `range()` helper used `start >= end` which caused it to return an empty array when `start === end` (i.e. totalPages=1, currentPage=1). This meant no page number was rendered between the prev/next arrows.

Changed the guard to `start > end` so that `range(1, 1)` correctly returns `[1]`, and updated the JSDoc and tests accordingly.

**Before:** Pagination with `totalPages={1}` shows no page number
**After:** Shows `1` between the arrows as expected

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected pagination range handling so equal start and end values now return a single-item range.
  - Invalid ranges are limited to cases where the start value is greater than the end value.


Abdellox:
Withdrawing this PR — flowbite-react has had no commits on main since June 2026 and it's no longer an active project to contribute to. The fix is tested and ready if the project becomes active again.

PR #1691, opened 2026-09-06
fix(RangeSlider): vertically center slider in its container

### LINKED ISSUE #1202: Range slider is not vertically centered

Abdellox:
Fixes #1202

The root wrapper used `flex` without `items-center`, causing the range input to sit at the bottom of the flex container instead of being vertically centered. This made it impossible to align the RangeSlider with other components using `flex` + `items-center`.

Added `items-center` to the root base Tailwind class to fix vertical alignment.

**Before:** Slider touches the bottom of its container box
**After:** Slider is vertically centered — works correctly with sibling flex items

## Summary by CodeRabbit

- **Style**
  - Improved range slider layout by vertically centering its content within the slider container.


Abdellox:
Withdrawing this PR — flowbite-react has had no commits on main since June 2026 and it's no longer an active project to contribute to. The fix itself is tested and ready if the project becomes active again.

